PR #22, opened 2021-09-16
Move span type up in snapshot output

Kyle-Verhoog:


ddapm_test_agent/snapshot.py
```diff
@@ -253,6 +253,35 @@ def snapshot(
...
+            if k in ["meta", "metrics"]:
+                # Sort the meta and metrics dictionaries alphanumerically
+                d[k] = OrderedDict(sorted(s[k].items(), key=operator.itemgetter(0)))  # type: ignore
+            else:
+                d[k] = s[k]  # type: ignore
+
+    # Add the rest of the attributes in alphanumeric order.
+    for k in sorted(set(s.keys()) - set(order)):
+        d[k] = s[k]  # type: ignore
+    return d  # type: ignore
```
Kyle-Verhoog, line 282: bunch of ignores because it's hard to go between Dict and OrderedDict 😕. I attempted to more generally type spans with Mapping but methods like `copy()` don't exist on Mapping so that would require quite a bit of work to refactor.

PR #64, opened 2022-04-07
Fix HTTPStatusCode and snapshot ordering

Kyle-Verhoog:


ddapm_test_agent/tracestats_snapshot.py
```diff
@@ -21,9 +21,17 @@ def _normalize_statsbuckets(buckets: List[StatsBucket]) -> List[StatsBucket]:
 
     # Sort aggr for a bucket alphanumerically
     for bucket in normed_buckets:
-        # Sort aggrs by name then resource then hits
+        # Sort aggrs since order does not matter.
```
brettlangdon, line 24: this comment doesn't make sense...?

"Sort them because the order doesn't matter"?

Or is this meant to be "Sort them because the order we receive them doesn't matter"? ... ?

brettlangdon (approved):

PR #71, opened 2022-05-09
Generate better snapshot error messages

Kyle-Verhoog:
Be more verbose about received/expected traces rather than just spitting
out a normalized trace id.


Kyle-Verhoog:
Don't really know if this merits a release note and/or it could be included in a patch release 🤔

brettlangdon:
> Don't really know if this merits a release note and/or it could be included in a patch release 🤔

Seems like a feature to me ?

PR #75, opened 2022-05-12
Build docker image for ARM64 using buildx

### LINKED ISSUE #74: Publish ARM64 docker images
I'd like to use the mock agent on ARM64 as well as x64. AFACIT, there's no reason this _shouldn't_ be possible (the `python:3.9` base image supports it) it's just a bit of a pain as you need to use `buildx`. [I've done this on Windows before](https://andrewlock.net/creating-multi-arch-docker-images-for-arm64-from-windows/) where it's pretty easy, but it should be possible to do here too I think. 

The only downside is that the [VaultVulp/gp-docker-action](https://github.com/VaultVulp/gp-docker-action) action you're using at the moment doesn't seem to support it. There is [an official image for buildx from Docker here](https://github.com/marketplace/actions/docker-setup-buildx) which should be a good starting point.

I'm happy to do the grunt work to replace it if you're happy with that?

andrewlock:
This updates the build to use QEMU and buildx to build multi-arch images. I'm only building the amd64 and arm64 versions atm, but we could easily add more if there's demand for it (e.g. `linux/386`, `linux/arm/v7`). Didn't seem worth doing now as just adds to the build times

I have tested most of the pieces of this, the only part I can't safely test is the actual final push. You can see a demo run (where we build but don't push the image) [here](https://github.com/DataDog/dd-apm-test-agent/runs/6403683552?check_suite_focus=true).

Fixes #74


.github/workflows/docker.yml
```diff
@@ -5,28 +5,43 @@ on:
...
+    - name: Login to Docker
+      run: docker login -u publisher -p ${{ secrets.GITHUB_TOKEN }} ghcr.io
+
+    - name: Docker Build
+      uses: docker/build-push-action@v3
+      with:
+        push: true
+        tags: ${{ steps.set_names.outputs.image_name }}
+        platforms: "linux/amd64,linux/arm64/v8"
+        context: .
```
Kyle-Verhoog, line 47:
```suggestion
        context: .
 
```

.github/workflows/docker.yml
```diff
@@ -5,28 +5,44 @@ on:
...
+
```
Kyle-Verhoog, line 48:
```suggestion
```

PR #77, opened 2022-05-27
feat: add support for distributed traces

majorgreys:
Add support for distributed traces where an instrumented service sends a trace chunk where the root span has a parent not in the trace chunk.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,7 @@ The traces are normalized and output in JSON to a file. The following transforma
 
 - Trace ids are overwritten to match the order in which the traces were received.
 - Span ids are overwritten to be the DFS order of the spans in the trace tree.
+- Parent ids are overwritten using the normalized span ids. However, if the parent is not a span in the trace, the parent id is not overwritten. This is necessary for handling distributed traces where all spans are not sent to the same agent.
 - Span attributes are ordered to be more human-readable, with the important attributes being listed first.
 - Span attributes are otherwise ordered alphanumerically.
 - The span meta and metrics maps if empty are excluded.
--- a/ddapm_test_agent/trace.py
+++ b/ddapm_test_agent/trace.py
@@ -232,11 +232,30 @@ def copy_trace(t: Trace) -> Trace:
 
 def root_span(t: Trace) -> Span:
     """Return the root span of the trace."""
+    # Follow approach used in Datadog Agent: https://github.com/DataDog/datadog-agent/blob/927f9ca9acf7983b72a4bfbdd7a69132e1da8501/pkg/trace/traceutil/trace.go#L53
+
+    if len(t) == 0:
+        raise ValueError("empty trace: %s" % t)
+
+    # common case optimization to check for span where parent_id is either not
+    # set or set to 0
     for s in t:
         if "parent_id" not in s or s["parent_id"] is None or s["parent_id"] == 0:
             return s
 
-    raise ValueError("root span not found in trace: %s" % t)
+    # collect root spans as those with parents that are not themselves spans in trace
+    span_ids = set(s["span_id"] for s in t)
+    roots = {
+        s["parent_id"]: s
+        for s in t
+        if "parent_id" in s and s["parent_id"] not in span_ids
+    }
+
+    if len(roots) != 1:
+        raise ValueError("single root span not found in trace (n=%d): %s" % (len(t), t))
+
+    # return any root candidate
+    return roots.popitem()[1]
 
 
 def trace_id(t: Trace) -> TraceId:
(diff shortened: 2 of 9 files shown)


ddapm_test_agent/trace.py
```diff
@@ -232,11 +232,30 @@ def copy_trace(t: Trace) -> Trace:
...
+    # collect root spans as those with parents that are not themselves spans in trace
+    span_ids = set(s["span_id"] for s in t)
+    roots = {
+        s["parent_id"]: s
+        for s in t
+        if "parent_id" in s and s["parent_id"] not in span_ids
+    }
+
+    if len(roots) != 1:
+        raise ValueError("single root span not found in trace: %s" % t)
```
Kyle-Verhoog, line 255: probably worth including the number of root spans in this message (and their operation names)

Kyle-Verhoog (commented):
We should have an integration snapshot test with a distributed trace (shouldn't be too bad to add to `test_snapshot_integration.py`). 

We'll have to figure out if we should leave the parent id as it is received or if we normalize it at all. I also suspect the snapshot logic assumes that all span ids have corresponding spans in the trace

majorgreys:
Blocked on https://github.com/DataDog/sketches-py/pull/55

setup.py
```diff
@@ -29,6 +29,7 @@
     install_requires=[
         "aiohttp",
         "ddsketch",
+        "protobuf<4.21.0",
```
Kyle-Verhoog, line 32:
```suggestion
```

tests/integration_snapshots/test_trace_distributed_propagated.json
```diff
@@ -0,0 +1,29 @@
+[[
+  {
+    "name": "root",
+    "service": null,
+    "resource": "root",
+    "trace_id": 0,
+    "span_id": 1,
+    "parent_id": 5678,
```
Kyle-Verhoog, line 8: 👍

Kyle-Verhoog (commented):
can we just add a bullet point here: https://github.com/DataDog/dd-apm-test-agent#snapshot-output

to state how distributed traces (when some spans aren't sent) are handled?

once that's done then LGTM 🙌

ddapm_test_agent/trace.py
```diff
@@ -232,11 +232,30 @@ def copy_trace(t: Trace) -> Trace:
...
+        raise ValueError("empty trace: %s" % t)
+
+    # common case optimization to check for span where parent_id is either not
+    # set or set to 0
     for s in t:
         if "parent_id" not in s or s["parent_id"] is None or s["parent_id"] == 0:
             return s
 
-    raise ValueError("root span not found in trace: %s" % t)
+    # collect root spans as those with parents that are not themselves spans in trace
```
brettlangdon, line 246: with partial flushing, are we able to have multiple "root" spans where the parent is not in the trace?

```
| ------------------ A1  ----------------- |
| --- B1 --- | | --- B2 --- | | --- B3 --- |
  | -- C1 -- |   | -- C2 -- |   | -- C3 -- |
```

If we take the scenario where partial flushing is enabled with a min span value of `4`.

That would mean when B2 finishes, then B1, C1, B2, C2 would get flushed together in a payload. Then you have multiple "root" spans where the parent is not in the trace.

I know this is likely mostly a very specific edge case, but still possible.

brettlangdon: I suppose this would be true with a partial flush min span value of 3, 4, 5, or 6.

Kyle-Verhoog: we could quite easily add a test case for this scenario

Kyle-Verhoog: I think today apm assumes that "trace-level" data has to be on the first span of the payload in the event that there are multiple "roots" in the payload.

Kyle-Verhoog: Created an issue to track the work as I don't think it needs to block @majorgreys or this PR: https://github.com/DataDog/dd-apm-test-agent/issues/78

README.md
```diff
@@ -114,6 +114,7 @@ The traces are normalized and output in JSON to a file. The following transforma
 
 - Trace ids are overwritten to match the order in which the traces were received.
 - Span ids are overwritten to be the DFS order of the spans in the trace tree.
+- Parent ids are overwritten using the normalized span ids. However, if the parent is not a span in the trace, the parent id is not overwritten. This is necessary for handling distributed traces where all spans are not sent to the same agent.
```
Kyle-Verhoog, line 117: 👌 great description!

Kyle-Verhoog (approved):

brettlangdon (approved):

PR #83, opened 2022-06-13
Fix trace_content_length check

Kyle-Verhoog:
The check assumed that the header would always be included. Some
clients, particularly the go client, don't include this header.

### DIFF
--- a/ddapm_test_agent/agent.py
+++ b/ddapm_test_agent/agent.py
@@ -225,10 +225,7 @@ class Agent:
         with CheckTrace.add_frame("headers") as f:
             f.add_item(pprint.pformat(dict(request.headers)))
             checks.check("meta_tracer_version_header", headers=dict(request.headers))
-            checks.check(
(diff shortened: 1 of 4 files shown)


Kyle-Verhoog:
I attempted to add a regression test but the aiohttp client annoyingly doesn't allow the `Content-Length` header to be skipped

https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientSession

```
skip_auto_headers –

set of headers for which autogeneration should be skipped.

aiohttp autogenerates headers like User-Agent or Content-Type if these headers are not explicitly passed. Using skip_auto_headers parameter allows to skip that generation. Note that Content-Length autogeneration can’t be skipped.

Iterable of [str](https://docs.python.org/3/library/stdtypes.html#str) or [istr](https://multidict.readthedocs.io/en/stable/multidict.html#multidict.istr) (optional)
```

PR #85, opened 2022-06-15
Add /info endpoint support

Kyle-Verhoog:
The info endpoint exists to allow clients to detect what features are
available in the agent. The go tracer specifically will only enable
trace stats computation if it detects that the agent supports the
feature.

### DIFF
--- a/ddapm_test_agent/agent.py
+++ b/ddapm_test_agent/agent.py
@@ -215,6 +215,20 @@ class Agent:
         )
         return web.HTTPOk()
 
+    async def handle_info(self, request: Request) -> web.Response:
+        return web.json_response(
+            {
+                "version": "test",
+                "endpoints": [
+                    "/v0.4/traces",
+                    "/v0.5/traces",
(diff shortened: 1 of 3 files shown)


ddapm_test_agent/agent.py
```diff
@@ -215,6 +215,20 @@ async def handle_v06_tracestats(self, request: Request) -> web.Response:
         )
         return web.HTTPOk()
 
+    async def handle_info(self, request: Request) -> web.Response:
+        return web.json_response(
```
brettlangdon, line 219: hmm, weird to have this as a mock?

Kyle-Verhoog: 🤷‍♂️ I think I'm ok with it, ensures that it doesn't change unexpectedly

Kyle-Verhoog: mmm it's going to be fundamentally different since the agents support different features

brettlangdon (approved):

PR #89, opened 2022-07-01
Add `metrics.process_id` to default ignored attributes

### LINKED ISSUE #88: Should `metrics.process_id` be ignored by default?
I've just started running some snapshot tests with services running in Windows, and the snapshots are currently failing because of differences in `metrics.process_id`. 

I can add that to the `SNAPSHOT_IGNORED_ATTRS` list, but should that be ignored by default? I see `metrics.system.pid` is already ignored so I guess a secondary question is: which is the "correct" attribute 😅

andrewlock:
Fixes #88

### DIFF
--- a/ddapm_test_agent/trace_snapshot.py
+++ b/ddapm_test_agent/trace_snapshot.py
@@ -29,9 +29,7 @@ from .trace import trace_id as get_trace_id
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


Kyle-Verhoog (approved):
👍 👍

brettlangdon (approved):

PR #90, opened 2022-07-15
Add ClientDropP0s feature to info endpoint

Kyle-Verhoog:
This feature goes with the /v0.6/stats (trace stats) implementation.

### DIFF
--- a/ddapm_test_agent/agent.py
+++ b/ddapm_test_agent/agent.py
@@ -227,6 +227,7 @@ class Agent:
                 ],
                 "feature_flags": [],
                 "config": {},
+                "client_drop_p0s": True,
             }
         )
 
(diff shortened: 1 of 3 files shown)


ddapm_test_agent/agent.py
```diff
@@ -227,6 +227,7 @@ async def handle_info(self, request: Request) -> web.Response:
                 ],
                 "feature_flags": [],
                 "config": {},
+                "client_drop_p0s": True,
```
brettlangdon, line 230: should this response value always be semi-fixed based on recent version of the agent?

should we have tests that validate we return the same value as the agent?

Kyle-Verhoog: AFAICT it's just `true` if the stats endpoint is implemented... idk why there's a separate option for it if that is the case though 🤔

PR #91, opened 2022-07-21
feat(telemetry): add initial support for capturing telemetry event requests

brettlangdon:


ddapm_test_agent/apmtelemetry.py
```diff
@@ -0,0 +1,12 @@
+import json
+from typing import Any
+from typing import cast
+
+
+TelemetryEvent = dict[str, Any]
+
+
+def decode_v2(data: bytes) -> TelemetryEvent:
```
Kyle-Verhoog, line 9:
```suggestion
def v2_decode(data: bytes) -> TelemetryEvent:
```

nit

ddapm_test_agent/apmtelemetry.py
```diff
@@ -0,0 +1,12 @@
+import json
+from typing import Any
+from typing import cast
+
+
+TelemetryEvent = dict[str, Any]
```
Kyle-Verhoog, line 6: I think this breaks Python <3.9 compatibility (forget specifically which version introduced the type thing)

```suggestion
TelemetryEvent = Dict[str, Any]
```

Kyle-Verhoog: Can we/should we be more specific about the type?

ddapm_test_agent/agent.py
```diff
@@ -21,6 +21,8 @@
 from . import _get_version
 from . import trace_snapshot
 from . import tracestats_snapshot
+from .apmtelemetry import TelemetryEvent
+from .apmtelemetry import decode_v2 as apmtelemetry_decode_v2
```
Kyle-Verhoog, line 25:
```suggestion
from .apmtelemetry import v2_decode as v2_apmtelemetry_decode
```
nit

ddapm_test_agent/agent.py
```diff
@@ -451,6 +500,27 @@ async def handle_test_traces(self, request: Request) -> web.Response:
...
+            "runtime_ids", request.headers.get("X-Datadog-Runtime-Ids", "")
+        )
+        if raw_runtime_ids:
+            runtime_ids = raw_runtime_ids.split(",")
+            events: List[TelemetryEvent] = []
+            for rid in runtime_ids:
+                try:
+                    events.extend(await self._apmtelemetry_by_runtime_id(rid))
+                except KeyError:
+                    events.append({})
```
Kyle-Verhoog, line 519: append an empty event? Wonder if we should fail instead?

Don't feel super strongly about one way or the other.

Kyle-Verhoog (commented):
mostly nits but lgtm once resolved

ddapm_test_agent/agent.py
brettlangdon: I was just trying to follow the pattern used for traces... I think?

I borrowed this idea, it is weird

ddapm_test_agent/apmtelemetry.py
brettlangdon: yes, I want to in the future, but this should be good enough for now.

we have an openapi spec for this endpoint, so we can/should add better validation

ddapm_test_agent/agent.py
```diff
@@ -509,14 +576,19 @@ def make_app(
             web.put("/v0.5/traces", agent.handle_v05_traces),
             web.post("/v0.6/stats", agent.handle_v06_tracestats),
             web.put("/v0.6/stats", agent.handle_v06_tracestats),
+            web.post(
```
Kyle-Verhoog, line 579: `put` is probably also supported but i'm fine to leave this for a follow up

ddapm_test_agent/apmtelemetry.py
```diff
@@ -0,0 +1,13 @@
+import json
+from typing import Any
+from typing import Dict
+from typing import cast
+
+
+TelemetryEvent = Dict[str, Any]
```
Kyle-Verhoog, line 7: would be really cool to flesh out this schema, but all good for now!

tests/test_agent.py
```diff
@@ -67,3 +67,40 @@ async def test_info(agent):
         "config": {},
         "client_drop_p0s": True,
     }
+
+
+async def test_apmtelemetry(
```
Kyle-Verhoog, line 72: we could probably add an integration test with `ddtrace` (manually enable the telemetry writer) but can follow up with this.

Kyle-Verhoog (approved):

ddapm_test_agent/agent.py
Kyle-Verhoog: ah, yeah that is weird 🤷‍♂️

ddapm_test_agent/agent.py
brettlangdon: The API schema/docs only say `POST`, so I'd rather leave that as-is.

ddapm_test_agent/apmtelemetry.py
brettlangdon: yeah, I haven't decided yet, but since we have an OpenAPI schema, we can play with how we introduce that. is it a check or part of this decoding?

Kyle-Verhoog (approved):
🙌

PR #93, opened 2022-08-02
Set 722 file permissions on UDS socket

Kyle-Verhoog:
The permissions of the socket created by aiohttp aren't as permissive as
the ones set by the real agent. (see
https://github.com/aio-libs/aiohttp/issues/4155)

from @devinsba:

test agent:

```
cnb@my-app:/var/run/datadog$ ls -la
total 4
drwxr-xr-x 2 root root   60 Aug  1 21:24 .
drwxr-xr-x 1 root root 4096 Aug  1 21:24 ..
srwxr-xr-x 1 root root    0 Aug  1 21:24 apm.socket
```

real agent:

```
cnb@my-app:/var/run/datadog$ ls -la
total 4
drwxr-xr-x 2 root root   80 Aug  1 21:30 .
drwxr-xr-x 1 root root 4096 Aug  1 21:31 ..
srwx-w--w- 1 root root    0 Aug  1 21:30 apm.socket
```

aiohttp doesn't seem to have a configurable way
to set these permissions but it does let us create and
manage a socket ourselves, so let's try that.

### DIFF
--- a/ddapm_test_agent/agent.py
+++ b/ddapm_test_agent/agent.py
@@ -1,10 +1,12 @@
 import argparse
+import atexit
 import base64
 from collections import OrderedDict
 import json
 import logging
 import os
 import pprint
+import socket
 import sys
 from typing import Awaitable
 from typing import Callable
@@ -722,6 +724,13 @@ def main(args: Optional[List[str]] = None) -> None:
         print(_get_version())
         sys.exit(0)
 
+    apm_sock: Optional[socket.socket] = None
+    if parsed_args.trace_uds_socket is not None:
+        apm_sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+        apm_sock.bind(parsed_args.trace_uds_socket)
+        os.chmod(parsed_args.trace_uds_socket, 0o722)
+        atexit.register(lambda: os.unlink(parsed_args.trace_uds_socket))
+
     if parsed_args.trace_request_delay is not None:
         log.info(
             "Trace request stall seconds setting set to %r.",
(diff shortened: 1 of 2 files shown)


ddapm_test_agent/agent.py
```diff
@@ -722,6 +724,13 @@ def main(args: Optional[List[str]] = None) -> None:
         print(_get_version())
         sys.exit(0)
 
+    apm_sock: Optional[socket.socket]
+    if parsed_args.trace_uds_socket is not None:
+        apm_sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+        apm_sock.bind(parsed_args.trace_uds_socket)
+        os.chmod(parsed_args.trace_uds_socket, 0o777)
```
devinsba, line 731: 722 is probably fine since that is what the agent does but I understand wanting to be permissive

ddapm_test_agent/agent.py
```diff
@@ -722,6 +724,13 @@ def main(args: Optional[List[str]] = None) -> None:
         print(_get_version())
         sys.exit(0)
 
+    apm_sock: Optional[socket.socket] = None
+    if parsed_args.trace_uds_socket is not None:
+        apm_sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+        apm_sock.bind(parsed_args.trace_uds_socket)
+        os.chmod(parsed_args.trace_uds_socket, 0o777)
+        atexit.register(lambda: os.unlink(parsed_args.trace_uds_socket))
```
devinsba, line 732: This is cool, this is the part I was missing this morning

devinsba (approved):

ddapm_test_agent/agent.py
Kyle-Verhoog: good call, let's just mirror what the agent does

PR #100, opened 2022-10-24
Add `metrics.system.process_id` to ignored tags

Kyle-Verhoog:
This tag is replacing the system.pid tag.


Kyle-Verhoog:
cc @wconti27

PR #114, opened 2023-03-29
[PHP] changes needed for integrating Test Agent with PHP Tracer

wconti27:
This PR is for changes for necessary for the PHP tracer integration.

PR for small changes needed for integration the APM Test agent within each tracer's test suite / CI. This PR will hold any minor changes necessary when trying to first integrate the Test Agent into each additional tracer's CI test suite. Integrations by repo so far:

- [x] Python - Already used
- [ ] PHP - PR adding Test Agent in Review - Small change here to ensure traces are always proxied even if there's an error.
- [ ] NodeJS - In Progress - No necessary major changes expected 
- [ ] .NET - Confirm Test Agent fully In Use?
- [ ] Ruby
- [ ] Go
- [ ] Java - Challenging :( - Will need another PR


Kyle-Verhoog (changes requested):
these changes need to be split up and justified, this is far too many changes introduced without description

ddapm_test_agent/trace.py
```diff
@@ -90,8 +107,8 @@ def verify_span(d: Any) -> Span:
         for attr in required_attrs:
             assert attr in d, f"'{attr}' required in span"
         NoneType = type(None)
-        assert isinstance(d["span_id"], int)
-        assert isinstance(d["trace_id"], int)
+        assert isinstance(d["span_id"], (float, int))
```
Kyle-Verhoog, line 110: span_id should not be a float

riotfile.py
```diff
@@ -28,11 +28,11 @@
                 ),
                 Venv(
                     name="fmt",
-                    command="isort . && black .",
+                    command="isort --profile black . && black .",
```
Kyle-Verhoog, line 31: this change should be made separately

ddapm_test_agent/trace.py
```diff
@@ -273,7 +293,12 @@ def set_metric_tag(s: Span, k: str, v: MetricType) -> Span:
 
 def decode_v04(content_type: str, data: bytes) -> v04TracePayload:
```
Kyle-Verhoog, line 294: can we do these changes to the decode error logging separately. I'd like to have tests supporting it

tests/test_checks.py
```diff
@@ -19,20 +19,20 @@ async def test_reference(
...
-        "/v0.4/traces",
-        headers={
-            "content-type": "application/msgpack",
-            "x-datadog-trace-count": "1",
-            "datadog-meta-tracer-version": "v0.1",
-        },
-        data=v04_reference_http_trace_payload_data,
-    )
-    assert resp.status == 200, await resp.text()
+# async def test_reference_case_insensitive(
```
Kyle-Verhoog, line 22: ?

ddapm_test_agent/agent.py
```diff
@@ -51,6 +51,9 @@
 log = logging.getLogger(__name__)
 
 
+AGENT_PROXY_HOST = os.environ.get("PROXY_HOST", "127.0.0.1")
```
Kyle-Verhoog, line 54: this should be set like the rest of the arguments/settings:

https://github.com/DataDog/dd-apm-test-agent/blob/232a2c6b92a2b032357741895a49c92c92dce860/ddapm_test_agent/agent.py#L725-L730

Kyle-Verhoog (changes requested):

ddapm_test_agent/trace.py
wconti27: done

tests/test_checks.py
wconti27: wasn't passing earlier for some reason

ddapm_test_agent/agent.py
```diff
@@ -330,53 +333,70 @@ async def _handle_traces(self, request: Request, version: Literal["v0.4", "v0.5"
         token = request["session_token"]
         checks: Checks = request.app["checks"]
 
-        await checks.check("trace_stall", headers=request.headers, request=request)
+        await checks.check("trace_stall", headers=dict(request.headers), request=request)
```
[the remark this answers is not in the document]
wconti27: removed all the dict uses

ddapm_test_agent/agent.py
wconti27: removed this argument, have no need for it for PHP

ddapm_test_agent/trace.py
wconti27: fine by me, I removed it

Dockerfile
```diff
@@ -1,13 +1,25 @@
-FROM python:3.11
```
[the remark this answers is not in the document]
wconti27: removed them

ddapm_test_agent/agent.py
```diff
@@ -333,50 +333,74 @@ async def _handle_traces(self, request: Request, version: Literal["v0.4", "v0.5"
...
+                        num_traces=len(traces),
+                    )
+            except Exception as e:
+                log.error(e)
 
         agent_url = request.app["agent_url"]
         if agent_url:
             log.info("Forwarding request to agent at %r", agent_url)
+
+            headers = {
```
Kyle-Verhoog, line 374: we shouldn't be using a `dict` for this as there can be duplicate http headers. To ensure we accurately pass the request along we should avoid removing duplicate headers.

Instead we should be using a list

ddapm_test_agent/agent.py
```diff
@@ -333,50 +333,74 @@ async def _handle_traces(self, request: Request, version: Literal["v0.4", "v0.5"
...
+                "Content-Type": "application/msgpack",
```
Kyle-Verhoog, line 375: why are we setting the content type? the library should be setting this

```suggestion
```

ddapm_test_agent/agent.py
```diff
@@ -333,50 +333,74 @@ async def _handle_traces(self, request: Request, version: Literal["v0.4", "v0.5"
...
-                    headers=request.headers,
+                    headers=headers,
                     data=self._request_data(request),
                 ) as resp:
                     assert resp.status == 200
-                    data = await resp.json()
-                    log.info("Got response %r from agent", data)
-                    return web.json_response(data=data)
+
+                    if "text/html" in resp.content_type:
```
Kyle-Verhoog, line 386: This is a good change. Can we add a release note for it?

ddapm_test_agent/trace.py
```diff
@@ -17,6 +18,9 @@
 from typing_extensions import TypedDict
 
 
+log = logging.getLogger(__name__)
```
Kyle-Verhoog, line 21:
```suggestion
```

this module provides fuctional utility methods, there shouldn't be any logging done. If error information is required then the exceptions can be caught by the caller of these functions and logged there.

ddapm_test_agent/trace.py
```diff
@@ -115,13 +132,16 @@ def verify_span(d: Any) -> Span:
                 assert isinstance(v, (int, float))
         return cast(Span, d)
     except AssertionError as e:
+        log.info(d)
```
Kyle-Verhoog, line 135: logging should be left to the caller of this function

ddapm_test_agent/trace.py
```diff
@@ -47,6 +51,19 @@
 MetricType = Union[int, float]
 
 
+def convert_numbers(data):
```
Kyle-Verhoog, line 54: why do we need this method?

ddapm_test_agent/trace.py
```diff
@@ -47,6 +51,19 @@
...
+    if isinstance(data, list):
+        return [convert_numbers(item) for item in data]
+    elif isinstance(data, dict):
+        return {key: convert_numbers(value) for key, value in data.items() if key != "meta"}
```
Kyle-Verhoog, line 58: the only dict we'd expect that isn't `meta` would be `metrics`. Why do we need to convert the values in `metrics` to numeric types?

Kyle-Verhoog (changes requested):

ddapm_test_agent/agent.py
```diff
@@ -329,23 +331,42 @@ async def _handle_traces(self, request: Request, version: Literal["v0.4", "v0.5"
...
+        headers = CIMultiDict(request.headers)
 
-        await checks.check("trace_stall", headers=request.headers, request=request)
+        log.debug(f"New request: {request} with headers: {headers}")
+        log.debug(f"Request Data: {self._request_data(request)!r}")
+
+        await checks.check("trace_stall", headers=headers, request=request)
+
+        proxy_to_agent = True
+        if "Do-Not-Proxy-To-Agent" in headers:
```
Kyle-Verhoog, line 342:
```suggestion
        if "Datadog-Agent-Proxy-Enabled" in headers:
```

Can we document this here https://github.com/DataDog/dd-apm-test-agent#proxy

also add a release note

ideally this is its own change

ddapm_test_agent/agent.py
```diff
@@ -357,26 +378,52 @@ async def _handle_traces(self, request: Request, version: Literal["v0.4", "v0.5"
...
+                with CheckTrace.add_frame(f"payload ({len(traces)} traces)"):
+                    await checks.check(
+                        "trace_count_header",
+                        headers=headers,
+                        num_traces=len(traces),
+                    )
 
         agent_url = request.app["agent_url"]
-        if agent_url:
+        if agent_url and proxy_to_agent:
```
Kyle-Verhoog, line 390: let's move this proxying up so that we proxy right away to avoid having to handle errors in the test agent to get to this point

ddapm_test_agent/trace.py
```diff
@@ -271,11 +275,45 @@ def set_metric_tag(s: Span, k: str, v: MetricType) -> Span:
...
+    parsed_data = json.loads(json_string, object_hook=json_decoder)
+    return parsed_data
+
+
 def decode_v04(content_type: str, data: bytes) -> v04TracePayload:
     if content_type == "application/msgpack":
         payload = msgpack.unpackb(data)
     elif content_type == "application/json":
-        payload = json.loads(data)
+        payload = parse_json(data)
```
Kyle-Verhoog, line 316: can we introduce a check (default disabled) to enable this parsing. We don't want libraries that don't need this to regress to having this behaviour

ddapm_test_agent/agent.py
```diff
@@ -357,26 +378,52 @@ async def _handle_traces(self, request: Request, version: Literal["v0.4", "v0.5"
...
             log.info("Forwarding request to agent at %r", agent_url)
+
+            proxy_headers = {
```
Kyle-Verhoog, line 393: can we do an exclusion list of headers that we suspect the test agent is adding to requests rather than trying to pick out the ones we think are coming from the client library

ddapm_test_agent/agent.py
```diff
@@ -329,23 +331,42 @@ async def _handle_traces(self, request: Request, version: Literal["v0.4", "v0.5"
...
+                if version == "v0.4":
+                    traces = self._decode_v04_traces(request)
+                elif version == "v0.5":
+                    traces = self._decode_v05_traces(request)
+                log.info(
+                    "received trace for token %r payload with %r trace chunks",
+                    token,
+                    len(traces),
+                )
+            except TypeError as e:
```
Kyle-Verhoog, line 363: this can be removed when we introduce the check

ddapm_test_agent/agent.py
```diff
@@ -357,26 +378,52 @@ async def _handle_traces(self, request: Request, version: Literal["v0.4", "v0.5"
...
                     f"{agent_url}/v0.4/traces",
-                    headers=request.headers,
+                    headers=proxy_headers,
                     data=self._request_data(request),
                 ) as resp:
-                    assert resp.status == 200
-                    data = await resp.json()
-                    log.info("Got response %r from agent", data)
-                    return web.json_response(data=data)
+                    log.info("Got response from agent with status: %s", resp.status)
```
Kyle-Verhoog, line 403: can we pull this change into its own PR

PR #121, opened 2023-04-21
Change trace header handling

wconti27:
Change trace headers to not automatically use the Python client library to set headers such as content-length, which when combined with traces from the PHP tracer, resulted in content-types of "octet/stream", which is invalid. Changed the headers to use the already set content-type from the original request and then to keep all other remaining headers. The PR also adds an optional header that can be included in the trace-payload to prevent proxying to the end-agent. This is useful in cases where most requests received by the Test-Agent need to be proxied, with the exception of a few.

### DIFF
--- a/releasenotes/notes/change-trace-proxy-headers-f9fcbb4efc4fcb95.yaml
+++ b/releasenotes/notes/change-trace-proxy-headers-f9fcbb4efc4fcb95.yaml
@@ -2,5 +2,5 @@
 features:
   - |
     This change adds support for disabling / enabling proxying for a trace by sending 
-    the Datadog-Agent-Proxy-Enabled with a boolean value within a trace's headers. By default,
+    the X-Datadog-Agent-Proxy-Disabled with a boolean value within a trace's headers. By default,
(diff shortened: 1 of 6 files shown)


releasenotes/notes/change-trace-proxy-headers-f9fcbb4efc4fcb95.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    This change adds support for disabling / enabling proxying for a trace by sending 
+    the Datadog-Agent-Proxy-Enabled with a boolean value within a trace's headers. By default,
```
Kyle-Verhoog, line 5: Can we document this option in the README too?

Kyle-Verhoog (commented):
just one nit, otherwise LGTM!

releasenotes/notes/change-trace-proxy-headers-f9fcbb4efc4fcb95.yaml
wconti27: added

Kyle-Verhoog (approved):

PR #123, opened 2023-05-15
add request forwarding for telemetry

wconti27:
This PR adds two middlewares to the Test-Agent. The first middleware stores the request data and headers in Test-Agent memory. The second middleware forwards requests to an optional end destination (such as a real DD Agent) for any requests from the `/traces`, `/stats`, or `/telemetry` endpoints.

### DIFF
--- a/ddapm_test_agent/agent.py
+++ b/ddapm_test_agent/agent.py
@@ -637,7 +637,26 @@ class Agent:
     @middleware  # type: ignore
     async def store_request_middleware(self, request: Request, handler: _Handler) -> web.Response:
         await self._store_request(request)
-        # Call the original handler
+
+        headers = CIMultiDict(request.headers)
+
+        if "X-Datadog-Trace-Env-Variables" in headers:
+            var_string = headers.pop("X-Datadog-Trace-Env-Variables")
+            env_vars = {
+                key.strip(): value.strip() for key, value in (pair.split("=") for pair in var_string.split(","))
+            }
+            request["_dd_trace_env_variables"] = env_vars
+
+        if "X-Datadog-Agent-Proxy-Disabled" in headers:
+            request["_proxy_to_agent"] = headers.pop("X-Datadog-Agent-Proxy-Disabled").lower() != "true"
+
+        if "X-Datadog-Proxy-Port" in headers:
+            port = headers.pop("X-Datadog-Proxy-Port")
+            request.app["agent_url"] = update_trace_agent_port(request.app["agent_url"], new_port=port)
+            log.info("Found port in headers, new trace agent URL is: {}".format(request.app["agent_url"]))
+
+        request["_headers"] = headers
+        # Call the original handlerx
         return await handler(request)
 
     @middleware  # type: ignore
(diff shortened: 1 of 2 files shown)


Kyle-Verhoog (changes requested):

ddapm_test_agent/agent.py
```diff
@@ -112,6 +112,48 @@ async def session_token_middleware(request: Request, handler: _Handler) -> web.R
     return await handler(request)
 
 
+async def _forward_request(request_data, headers, agent_url):
+    async with ClientSession() as session:
+        async with session.post(
+            f"{agent_url}/v0.4/traces",
```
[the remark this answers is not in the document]
wconti27: good catch

ddapm_test_agent/agent.py
```diff
@@ -652,6 +642,64 @@ async def handle_trace_analyze(self, request: Request) -> web.Response:
...
+        if "X-Datadog-Agent-Proxy-Disabled" in headers:
+            request["_proxy_to_agent"] = headers.pop("X-Datadog-Agent-Proxy-Disabled").lower() != "true"
+
+        if "X-Datadog-Proxy-Port" in headers:
+            port = headers.pop("X-Datadog-Proxy-Port")
+            request.app["agent_url"] = update_trace_agent_port(request.app["agent_url"], new_port=port)
+            log.info("Found port in headers, new trace agent URL is: {}".format(request.app["agent_url"]))
+
+        request["_headers"] = headers
+        # Call the original handlerx
```
Kyle-Verhoog, line 669:
```suggestion
        # Call the original handler
```

ddapm_test_agent/agent.py
```diff
@@ -652,6 +642,64 @@ async def handle_trace_analyze(self, request: Request) -> web.Response:
...
+
+        if "X-Datadog-Trace-Env-Variables" in headers:
+            var_string = headers.pop("X-Datadog-Trace-Env-Variables")
+            env_vars = {
+                key.strip(): value.strip() for key, value in (pair.split("=") for pair in var_string.split(","))
+            }
+            request["_dd_trace_env_variables"] = env_vars
+
+        if "X-Datadog-Agent-Proxy-Disabled" in headers:
+            request["_proxy_to_agent"] = headers.pop("X-Datadog-Agent-Proxy-Disabled").lower() != "true"
```
Kyle-Verhoog, line 661: this check should be in the proxy codepath, not the request storing

ddapm_test_agent/agent.py
```diff
@@ -652,6 +642,64 @@ async def handle_trace_analyze(self, request: Request) -> web.Response:
...
+        else:
+            # Call the original handler and do nothing
+            return await handler(request)
+
+    async def _forward_request_to_agent(self, request: Request, handler: _Handler) -> web.Response:
+        """Forward all requests to the agent_url if set."""
+        data = self._request_data(request)
+        headers = request["_headers"]
+        agent_url = request.app["agent_url"]
+        proxy_to_agent = request.get("_proxy_to_agent", True)
```
Kyle-Verhoog, line 686: why are we changing the condition for proxying to the agent? Before it was if there was an agent url detected (and then checking the header). 

```
        proxy_to_agent = agent_url != ""

        if "X-Datadog-Agent-Proxy-Disabled" in headers:
            proxy_to_agent = headers.pop("X-Datadog-Agent-Proxy-Disabled").lower() != "true"
```

can we maintain this functionality?

Kyle-Verhoog (commented):
This change needs a release note. Also it would be great if we could have tests for the request storing logic to ensure it works.

Otherwise, I really like the middleware implementation 👍

ddapm_test_agent/agent.py
wconti27: IMO I don't agree. I think centrally locating any header transformations / checks is useful. To get the same functionality as before, I would have to do this header check within each of the forward endpoints (traces, stats, telemetry).  I added the check to make sure agent_url isn't empty. Should be the same functionality as before.

ddapm_test_agent/agent.py
wconti27: This functionality is maintained

ddapm_test_agent/agent.py
Kyle-Verhoog: wouldn't it be centralized in the proxy codepath/middleware?

ddapm_test_agent/agent.py
```diff
@@ -134,6 +177,14 @@ def __init__(self):
         # Token to be used if running test cases synchronously
         self._requests: List[Request] = []
         self._rc_server = RemoteConfigServer()
+        self._trace_failures: List[str] = []
```
Kyle-Verhoog, line 180:
```suggestion
```

unrelated change?

Kyle-Verhoog (changes requested):

Kyle-Verhoog (approved):

PR #125, opened 2023-05-19
Add `SNAPSHOT_REMOVED_ATTRS` to remove attributes from spans in snapshots

PROFeNoM:
Add a new _optional_ configuration to allow fields to be removed from the spans. By default, no keys will be removed from the snapshots.

💭 The motivation behind this feature is to get rid of some fields that can be noisy in the snapshots (e.g., `start` or `duration`) which could create some massive diffs when regenerating the snapshots.

Note that since the removal of the field `start` would cause an error when sorting the spans by this field, `span_id` was used as a fallback. Feel free to tell me if this would cause an issue I didn't consider (this is the reason why both `start` and `span_id` shouldn't be removed from the snapshots).

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ Please refer to `ddapm-test-agent-fmt --help` for more information.
 - `DD_SUPPRESS_TRACE_PARSE_ERRORS` [`false`]: Set to `"True"` to disable span parse errors when decoding handled traces. When disabled, errors will not be thrown for
 metrics incorrectly placed within the meta field, or other type errors related to span tag formatting/types. Can also be set using the `--suppress-trace-parse-errors=True` option.
 
-- `SNAPSHOT_REMOVED_ATTRS` []: The attributes to remove from spans in snapshots. This is useful for removing attributes 
+- `SNAPSHOT_REMOVED_ATTRS` [`""`]: The attributes to remove from spans in snapshots. This is useful for removing attributes 
 that are not relevant to the test case. **Note that you shouldn't remove both `start` and `span_id` to allow span 
 ordering to be completed.**
 
--- a/README.md
+++ b/README.md
@@ -177,8 +177,8 @@ Please refer to `ddapm-test-agent-fmt --help` for more information.
 metrics incorrectly placed within the meta field, or other type errors related to span tag formatting/types. Can also be set using the `--suppress-trace-parse-errors=True` option.
 
 - `SNAPSHOT_REMOVED_ATTRS` [`""`]: The attributes to remove from spans in snapshots. This is useful for removing attributes 
-that are not relevant to the test case. **Note that you shouldn't remove both `start` and `span_id` to allow span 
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -176,6 +176,10 @@ Please refer to `ddapm-test-agent-fmt --help` for more information.
 - `DD_SUPPRESS_TRACE_PARSE_ERRORS` [`false`]: Set to `"True"` to disable span parse errors when decoding handled traces. When disabled, errors will not be thrown for
 metrics incorrectly placed within the meta field, or other type errors related to span tag formatting/types. Can also be set using the `--suppress-trace-parse-errors=True` option.
 
+- `SNAPSHOT_REMOVED_ATTRS` []: The attributes to remove from spans in snapshots. This is useful for removing attributes
```
Kyle-Verhoog, line 179:
```suggestion
- `SNAPSHOT_REMOVED_ATTRS` [`""`]: The attributes to remove from spans in snapshots. This is useful for removing attributes 
```

ddapm_test_agent/agent.py
```diff
@@ -469,6 +469,15 @@ async def handle_snapshot(self, request: Request) -> web.Response:
         span_ignores = list(default_span_ignores | overrides)
         log.info("using ignores %r", span_ignores)
 
+        # Get the span attributes that are to be removed for this snapshot.
+        default_span_removes: Set[str] = request.app["snapshot_removed_attrs"]
+        overrides = set(_parse_csv(request.url.query.get("removes", "")))
+        span_removes = list(default_span_removes | overrides)
+        log.info("using removes %r", span_removes)
+
+        if "start" in span_removes and "span_id" in span_removes:
```
Kyle-Verhoog, line 478:
```suggestion
        if "span_id" in span_removes:
```

I think we can simplify this condition to just `span_id`. The span id is normalized (based off the `start` attribute) in the snapshot. It is used to assert the position and order of each span in the trace which makes it pretty crucial for the snapshot testing. Its value should be stable so I don't see a use-case where someone would want to ignore it.

tests/test_snapshot.py
```diff
@@ -425,3 +426,193 @@ async def test_snapshot_tracestats(agent, tmp_path, snapshot_ci_mode, do_referen
 
         resp = await agent.get("/test/session/snapshot", params={"test_session_token": "test_case"})
         assert resp.status == 200, await resp.text()
+
```
Kyle-Verhoog, line 429: can we add one more test case ensuring that when `span_id` is set as a removed attribute that the request fails?

Kyle-Verhoog (commented):
Great stuff, thank you for the extensive test cases! 🙌 

Just a couple of pretty small nits and this is good to go.

PROFeNoM:
Ok, applied these 🔧  Tell me if anything else is needed 👍

Kyle-Verhoog (approved):
🙌 great, thanks!

PR #128, opened 2023-06-05
Use correct encoded length in remoteconfig response

mcculls:
The Java tracer checks the decoded length matches the expected length, so the test agent must return the real length rather than a hard-coded value of `24` - otherwise the config will be rejected by the Java tracer.

cf: https://github.com/DataDog/dd-trace-java/blob/v1.15.3/remote-config/src/main/java/datadog/remoteconfig/tuf/RemoteConfigResponse.java#L144


Kyle-Verhoog (approved):

Kyle-Verhoog:
This catches a bug in the Python implementation where the size was not checked at all.

PR #129, opened 2023-06-22
feat: improve trace check results endpoint

wconti27:
PR adds improvements to the `tests/trace check/results` endpoint. It breaks down the number of checks run by type of trace check, adds the ability to clear Trace Check results, and can return results for the number of passed, failed, and skipped checks.

### DIFF
--- a/ddapm_test_agent/agent.py
+++ b/ddapm_test_agent/agent.py
@@ -804,7 +804,7 @@ class Agent:
             token = request["session_token"]
 
             # update trace_check results
-            self._trace_check_results_by_check[token] = trace.get_results(self._trace_check_results_by_check[token])
+            trace.update_results(self._trace_check_results_by_check[token])
 
             # only save trace failures to memory if necessary
             msg = str(trace) + str(e)
@@ -817,7 +817,7 @@ class Agent:
         else:
             token = request["session_token"]
             # update trace_check results
-            self._trace_check_results_by_check[token] = trace.get_results(self._trace_check_results_by_check[token])
+            trace.update_results(self._trace_check_results_by_check[token])
             if trace.has_fails():
                 # only save trace failures to memory if necessary
                 pool_failures = request.app["pool_trace_check_failures"]
--- a/ddapm_test_agent/agent.py
+++ b/ddapm_test_agent/agent.py
@@ -10,6 +10,7 @@ import socket
 import sys
 from typing import Awaitable
 from typing import Callable
+from typing import Dict
 from typing import List
 from typing import Literal
 from typing import Optional
@@ -159,7 +160,8 @@ class Agent:
         # Token to be used if running test cases synchronously
         self._requests: List[Request] = []
         self._rc_server = RemoteConfigServer()
-        self._trace_failures: List[str] = []
+        self._trace_failures: Dict[str, List[str]] = {"default": []}
+        self._trace_check_results_by_check: Dict[str, Dict[str, int]] = {}
         self._forward_endpoints: List[str] = [
             "/v0.4/traces",
             "/v0.5/traces",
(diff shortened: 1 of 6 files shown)


ddapm_test_agent/agent.py
```diff
@@ -187,18 +190,84 @@ async def traces(self) -> TraceMap:
                     _traces[trace_id].append(s)
         return _traces
 
+    async def clear_trace_check_failures(self, request: Request) -> web.Response:
+        """Clear traces by session token provided."""
+        token = request["session_token"]
+        clear_all = "clear_all" in request.query and request.query["clear_all"].lower() == "true"
```
tabgok, line 196: should this be `request.query.lower()` ?

ddapm_test_agent/agent.py
```diff
@@ -187,18 +190,84 @@ async def traces(self) -> TraceMap:
...
+        """Return the Trace Check failures that occurred, if pooling is enabled,
+        returned as either a Text (by default) or JSON response.
+        """
+        token = request["session_token"]
+        return_all = "return_all" in request.query and request.query["return_all"].lower() == "true"
+
+        if return_all:
+            # check for whether to return all results
+            trace_check_failures = []
+            for token, f in self._trace_failures.items():
```
tabgok, line 222: nit: you can extend on self._trace_failures.values() instead of using items.

ddapm_test_agent/agent.py
```diff
@@ -187,18 +190,84 @@ async def traces(self) -> TraceMap:
...
+                trace_check_failures.extend(f)
+            n_failures = len(trace_check_failures)
+            log.info(f"{n_failures} Trace Failures Occurred in Total")
+        else:
+            # or return results by token
+            trace_check_failures = self._trace_failures.get(token, [])
+            n_failures = len(trace_check_failures)
+            log.info(f"{n_failures} Trace Failures Occurred for Token {token}")
+        if n_failures > 0:
+            if "use_json" in request.query and request.query["use_json"].lower() == "true":
```
tabgok, line 232: same question: `request.query.lower()` ?

ddapm_test_agent/agent.py
```diff
@@ -187,18 +190,84 @@ async def traces(self) -> TraceMap:
...
+                for check_trace, failure_message in trace_check_failures:
+                    msg += failure_message
+                return web.HTTPBadRequest(text=msg)
         else:
             return web.HTTPOk()
 
+    def get_trace_check_summary(self, request: Request) -> web.Response:
+        token = request["session_token"]
+        summary: Dict[str, Dict[str, int]] = {}
+        return_all = "return_all" in request.query and request.query["return_all"].lower() == "true"
```
tabgok, line 251: `.lower()` ?

ddapm_test_agent/agent.py
```diff
@@ -187,18 +190,84 @@ async def traces(self) -> TraceMap:
...
+                msg = f"APM Test Agent Validation failed with {n_failures} Trace Check failures.\n"
+                for check_trace, failure_message in trace_check_failures:
+                    msg += failure_message
+                return web.HTTPBadRequest(text=msg)
         else:
             return web.HTTPOk()
 
+    def get_trace_check_summary(self, request: Request) -> web.Response:
+        token = request["session_token"]
+        summary: Dict[str, Dict[str, int]] = {}
```
tabgok, line 250: nit: you can use cllections.defaultdict() to clear up some of the if/else statements below

ddapm_test_agent/checks.py
```diff
@@ -85,6 +109,31 @@ def frames_dfs(self) -> Generator[Tuple[CheckTraceFrame, int], None, None]:
     def has_fails(self) -> bool:
         return len([f for f in self.frames() if f.has_fails()]) > 0
 
+    def get_results(self, results: Dict[str, Dict[str, int]]) -> Dict[str, Dict[str, int]]:
+        for f in self.frames():
+            results = f.get_results(results)
```
tabgok, line 114: do you mean to overwrite results each time?

ddapm_test_agent/agent.py
wconti27: I need to document this but `clear_all` is an optional query string parameter that can be accessed by using `request.query["clear_all"]`. I used `.lower()` in case it was included as: `/summary?clear_all=True`

ddapm_test_agent/agent.py
wconti27: thanks! much cleaner now

ddapm_test_agent/checks.py
wconti27: yes, we want the trace check results to be updated after each request, which means looping through each frame to get the results from that specific frame.

tabgok: oh, I think I see.  

What you probably want here is `update_results(...)` .  Since dicts are passed by reference in Python, the results dict will automatically be updated.

wconti27: fixed

tabgok (approved):

PR #131, opened 2023-06-29
Add flag for enabling non-default checks

devinsba:
Adds a config to allow some languages to use checks that are not enabled by default. Unblocks #127

### DIFF
--- a/ddapm_test_agent/agent.py
+++ b/ddapm_test_agent/agent.py
@@ -743,6 +743,7 @@ class Agent:
 
 def make_app(
     disabled_checks: List[str],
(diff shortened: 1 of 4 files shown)


wconti27 (approved):
Looks good thanks!

Kyle-Verhoog (commented):
Can we please add tests + docs to the README as well

devinsba:
Working on tests, can I just modify the existing release note file to get that check to pass?

Kyle-Verhoog:
there is the `no-changelog` github label for when a release note isn't needed

PR #132, opened 2023-06-30
Add test for additional checks config

devinsba:


ddapm_test_agent/agent.py
```diff
@@ -402,6 +403,8 @@ async def _handle_traces(self, request: Request, version: Literal["v0.4", "v0.5"
             f.add_item(pprint.pformat(dict(headers)))
             await checks.check("meta_tracer_version_header", headers=headers)
             await checks.check("trace_content_length", headers=headers)
+            for additional_check in checks.additional:
+                await checks.check(additional_check, headers=headers)
```
devinsba, line 407: This seems brittle to me but I don't see a need to make this more robust yet since there is only a single check that relies on anything other than headers

devinsba: Although the check that @wconti27 wants to add would check the trace content so maybe I should rework it. Thoughts @Kyle-Verhoog?

README.md
```diff
@@ -154,6 +154,8 @@ Please refer to `ddapm-test-agent-fmt --help` for more information.
 
 - `DISABLED_CHECKS` [`""`]: Comma-separated values of checks to disable.
 
+- `ADDITIONAL_CHECKS` [`""`]: Additional checks to enable that are not enabled by default. Currently unused
```
Kyle-Verhoog, line 157: what does "currently unused" mean?

tests/test_checks.py
```diff
@@ -115,3 +121,39 @@ async def test_trace_stall(
...
+    name = "for_testing"
+    description = """check for testing"""
+    default_enabled = False
+
+    def check(self, headers: CIMultiDictProxy) -> None:
+        if "For-Testing" not in headers:
+            self.fail("For-Testing not found in headers")
+
+
+class TestChecks:
```
Kyle-Verhoog, line 136: nit: this doesn't need to be a class-based test, we're not using any features of the class

ddapm_test_agent/agent.py
```diff
@@ -801,7 +806,8 @@ def make_app(
             CheckTraceCountHeader,
             CheckTraceContentLength,
             CheckTraceStallAsync,
-        ],
+        ]
+        + additional_check_classes,
```
Kyle-Verhoog, line 810: having 3 different ways to specify or not specify checks is getting a bit complex IMO. I wonder if we should just have the user of the test agent specify the checks they want and avoid having any defaults.

WDYT @wconti27 @devinsba?

wconti27: IMO I'm okay with this approach as it solves all requirements and I don't see needing any further complexity in the future. Having the user specify every check name to run in an env variable will lead to a very long variable value once we have 10+ checks.

Kyle-Verhoog: I think the verbosity is a feature as it becomes obvious what checks are enabled and it's obvious how to add/remove them.

@wconti27 maybe a file configuration is better suited to help the readability issue?

wconti27: Good point, having to explicitly specify checks does seem a lot more straightforward. Given that we don't have too many checks right now, Im fine with adding the config file as a possible future issue/improvement and just using an ENV variable for now. Would be great if we can get this merged soon in order to unblock the new service naming checks for Java/Python!

tests/test_checks.py
devinsba: OK. I wasn't sure if I could overwrite a fixture from a function based test or not

ddapm_test_agent/agent.py
devinsba: I'll rework it to make it completely opt in. I will create a PR to revert the previous PR and open a new one that reworks the whole thing

devinsba:
Closing in favor of opt-in checks list in a new PR

PR #155, opened 2023-12-28
Fix handling of traceflare requests

kevingosse:
Fixing a few errors in the way tracer flare requests are handled by the test agent.

- The parsing was failing because the MultipartReader was given an instance of asyncio.StreamReader instead of aiohttp.StreamReader
- When a parsing error occurred, it was not returned to the client


mcculls (approved):
LGTM, thanks

(note the tracer-flare sent by the Java tracer can be parsed by just using `asyncio.StreamReader` - I went with using that instead of `aiohttp.StreamReader` because of an issue getting it to use the recorded bytes, but it looks like setting the limit in the `MultipartReader` bypasses that issue so it's nice to remove that hack)

PR #156, opened 2024-01-09
logging: log received and expected if reasonable when len() mismatch

ZStriker19:
It's often very helpful to know the actual difference between received and expected snapshots when there's a length mismatch.

### DIFF
--- a/ddapm_test_agent/trace_snapshot.py
+++ b/ddapm_test_agent/trace_snapshot.py
@@ -262,15 +262,12 @@ def _compare_traces(expected: Trace, received: Trace, ignored: Set[str]) -> None
 
     The given traces are assumed to be in BFS order.
     """
-    if len(received) > len(expected):
+    if (len(received) != len(expected)):
         names = ["'%s'" % s["name"] for s in received[len(expected) - len(received) :]]
         raise AssertionError(
-            f"Received more spans ({len(received)}) than expected ({len(expected)}). Received unmatched spans: {', '.join(names)}"
-        )
-    elif len(expected) > len(received):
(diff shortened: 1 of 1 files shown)


ddapm_test_agent/trace_snapshot.py
```diff
@@ -262,15 +262,11 @@ def _compare_traces(expected: Trace, received: Trace, ignored: Set[str]) -> None
...
+    if len(received) != len(expected):
         raise AssertionError(
-            f"Received more spans ({len(received)}) than expected ({len(expected)}). Received unmatched spans: {', '.join(names)}"
-        )
-    elif len(expected) > len(received):
-        names = ["'%s'" % s["name"] for s in expected[len(received) - len(expected) :]]
-        raise AssertionError(
-            f"Received fewer spans ({len(received)}) than expected ({len(expected)}). Expected unmatched spans: {', '.join(names)}"
+            f"Received different number of spans ({len(received)}) than expected ({len(expected)})."
+            f"If spans<=20, will display received and expected.\nReceived: {received if (len(received) <= 20) else 'too long'}"
```
Kyle-Verhoog, line 268: How does this render? I imagine the raw python dicts might be difficult to read (which is why we had reduced it to just the span names before). What do you think about formatting them to make them more readable?

ddapm_test_agent/trace_snapshot.py
```diff
@@ -262,15 +262,11 @@ def _compare_traces(expected: Trace, received: Trace, ignored: Set[str]) -> None
...
+            f"\nExpected: {expected if (len(expected) <= 20) else 'too long'}"
```
Kyle-Verhoog, line 269: Could we resort to just the names in this case? Just so that there's at least some signal

PR #159, opened 2024-02-12
(feat): update session logic

tabgok:
Does the following:

- Clearing a session now clears traces only, instead of also deleting the session
- Requesting traces for a esssion will return a 404 Session Not found if no traces or session match a non-None session token
- Sessions now return session-less requests only up to the latest session marker

### DIFF
--- a/ddapm_test_agent/agent.py
+++ b/ddapm_test_agent/agent.py
@@ -63,6 +63,10 @@ from .tracestats import decode_v06 as tracestats_decode_v06
 from .tracestats import v06StatsPayload
 
 
+class NoSuchSessionException(Exception):
+    pass
+
+
 _Handler = Callable[[Request], Awaitable[web.Response]]
 
 
@@ -358,18 +362,34 @@ class Agent:
         """
         # Go backwards in the requests received gathering requests until
         # the /session-start request for the token is found.
-        reqs: List[Request] = []
-        for req in reversed(self._requests):
+        # Note that this may not return all associated traces, because some
+        # may be generated before the session-start call
+        session_reqs: List[Tuple[int, Request]] = []
+        sessionless_reqs: List[Tuple[int, Request]] = []
+        matched = token is None
+
+        for i, req in enumerate(reversed(self._requests)):
             if req.match_info.handler == self.handle_session_start:
-                if token is None or _session_token(req) == token:
+                if token is None:
+                    # If no token is specified, then we match the latest session
                     break
-                else:
-                    # The requests made were from a different manual session
-                    # so continue.
+                elif _session_token(req) == token:
+                    # If a token is specified and it matches, we've hit the start of our session
+                    matched = True
+                    break
+                elif _session_token(req) != token:
+                    # If a token is specified and it doesn't match, we've hit the start of a different session
+                    # So we reset the list of requests
+                    sessionless_reqs = []
                     continue
-            if _session_token(req) in [token, None]:
-                reqs.append(req)
-        return reqs
+            if _session_token(req) == token:
+                session_reqs.append((i, req))
+            elif _session_token(req) is None:
+                sessionless_reqs.append((i, req))
+
+        if not matched and not session_reqs:
+            raise NoSuchSessionException(f"No session found for token '{token}'")
+        return [x[1] for x in sorted(session_reqs + sessionless_reqs, key=lambda x: x[0])]
 
     async def _traces_from_request(self, req: Request) -> List[List[Span]]:
         """Return the trace from a trace request."""
@@ -814,7 +834,12 @@ class Agent:
 
     async def handle_session_traces(self, request: Request) -> web.Response:
         token = request["session_token"]
-        traces = await self._traces_by_session(token)
+        traces = []
+        try:
+            traces = await self._traces_by_session(token)
+        except NoSuchSessionException as e:
+            return web.HTTPNotFound(reason=str(e))
+
         return web.json_response(traces)
 
     async def handle_session_apmtelemetry(self, request: Request) -> web.Response:
@@ -908,14 +933,19 @@ class Agent:
                 if req.match_info.handler == self.handle_session_start:
                     if _session_token(req) == session_token:
                         in_token_sync_session = True
+                        continue  # Don't clear the session start
                     else:
                         in_token_sync_session = False
                 if in_token_sync_session:
                     setattr(req, "__delete", True)
 
-            # Filter out all the requests.
+            # Filter out all requests marked for deletion.
+            # Keep session starts.
             self._requests = [
-                r for r in self._requests if _session_token(r) != session_token and not hasattr(r, "__delete")
+                r
+                for r in self._requests
+                if (_session_token(r) != session_token or r.match_info.handler == self.handle_session_start)
+                and not hasattr(r, "__delete")
             ]
         else:
             self._requests = []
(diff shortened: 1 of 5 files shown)


ddapm_test_agent/trace_snapshot.py
```diff
@@ -30,7 +30,7 @@
 log = logging.getLogger(__name__)
 
 
-DEFAULT_SNAPSHOT_IGNORES = "span_id,trace_id,parent_id,duration,start,metrics.system.pid,metrics.system.process_id,metrics.process_id,meta.runtime-id,span_links.trace_id_high,meta.pathway.hash"
+DEFAULT_SNAPSHOT_IGNORES = "span_id,trace_id,parent_id,duration,start,metrics.system.pid,metrics.system.process_id,metrics.process_id,metrics._dd.tracer_kr,meta.runtime-id,span_links.trace_id_high,meta.pathway.hash"
```
Kyle-Verhoog, line 33: unrelated change?

tests/test_snapshot.py
```diff
@@ -431,6 +442,8 @@ async def test_snapshot_tracestats(agent, tmp_path, snapshot_ci_mode, do_referen
 
 @pytest.mark.parametrize("snapshot_removed_attrs", [{"start", "duration"}])
 async def test_removed_attributes(agent, tmp_path, snapshot_removed_attrs, do_reference_v04_http_trace):
+    resp = await agent.get("/test/session/start", params={"test_session_token": "test_case"})
```
Kyle-Verhoog, line 445: Is this request necessary? I worry that this will hurt the ergonomics of using the test agent if a test session token must be declared.

Kyle-Verhoog (commented):
Could you separate the 3 changes into 3 separate PRs? This would help make the changes more reviewable and make sure that we're testing each.

Just two questions for now pending the answer to the above question

ddapm_test_agent/trace_snapshot.py
tabgok: moved to another PR

tests/test_snapshot.py
tabgok: The behavior is now like this:

- Request with no token => get all traces
- Request with token + token declared => get all traces directly associated with that token + all unassociated requests until the start of the next session
- Request with token + token not declared => exception (no traces found)


These are being added because the tests are submitting a request with a token, but none has ever been declared

Kyle-Verhoog: So I think the first two points are fine. For the third do you mean there's an exception when querying back for that token? I would prefer to have

- Request with token + token not declared (no session start) => traces for that token

That way it isn't necessary for callers to have to make a separate request.

tabgok: > Request with token + token not declared (no session start) => traces for that token

Oh, whoops, my logic and my blurb don't match:

- Request with no token => get all traces
- Request with token + token declared => get all traces directly associated with that token + all unassociated requests until the start of the next session
- Request with token + token not declared + no requests/traces for session exist directly associated => exception (no traces found)

ddapm_test_agent/agent.py
```diff
@@ -345,18 +349,34 @@ def _requests_by_session(self, token: Optional[str]) -> List[Request]:
...
                     continue
-            if _session_token(req) in [token, None]:
-                reqs.append(req)
-        return reqs
+            if _session_token(req) == token:
+                session_reqs.append(req)
+            elif _session_token(req) is None:
+                sessionless_reqs.append(req)
+
+        if not matched and not session_reqs:
```
tabgok, line 377: Here's the "session created or we have at last one request with the session" logic

ddapm_test_agent/agent.py
```diff
@@ -895,14 +920,19 @@ async def handle_session_clear(self, request: Request) -> web.Response:
                 if req.match_info.handler == self.handle_session_start:
                     if _session_token(req) == session_token:
                         in_token_sync_session = True
+                        continue  # Don't clear the session start
```
Kyle-Verhoog, line 923: I wonder if we should add a way to maintain this functionality, `handle_session_delete` or something. However, I don't think there's actually a use-case for it so I'm fine leaving it until someone needs something like that.

tests/test_session.py
```diff
@@ -69,8 +69,7 @@ async def test_concurrent_session(
     assert resp.status == 200
     for token in ["test_case", "test_case2"]:
         resp = await agent.get("/test/session/traces", params={"test_session_token": token})
-        assert resp.status == 200
-        assert await resp.json() == []
+        assert resp.status == 404
```
Kyle-Verhoog, line 74: Can we add some regression tests to this file to cover the new functionality?

ddapm_test_agent/agent.py
```diff
@@ -345,18 +349,34 @@ def _requests_by_session(self, token: Optional[str]) -> List[Request]:
...
-                reqs.append(req)
-        return reqs
+            if _session_token(req) == token:
+                session_reqs.append(req)
+            elif _session_token(req) is None:
+                sessionless_reqs.append(req)
+
+        if not matched and not session_reqs:
+            raise NoSuchSessionException(f"No session found for token '{token}'")
+        return session_reqs + sessionless_reqs
```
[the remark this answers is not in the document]
tabgok: Added in a fix

ddapm_test_agent/agent.py
tabgok: Yeah, the functional behavior is preserved (no traces/snapshots will be returned), but there is a potential issue with memory space.  

I'd prefer to leave it for now

tests/test_session.py
tabgok: done

releasenotes/notes/clearing-sessions-now-only-removes-traces,-instead-of-sessions,-requsting-traces-will-now-return-a-404-error-for-TracesNotFound-if-a-session-is-specified-and-traces-are-not-found-6b0311673d8b37aa.yaml
```diff
@@ -0,0 +1,13 @@
+---
+features:
+  - |
+    Updates the logic for associating requsets with a session.  Prior to this change:
+    (1) non-existent tokens were permitted and returned associations with all requests, 
+    (2) existent tokens returned associations with all matching requests + all untokenized requests after the session was created
+
+    After this change:
+    (1) Requests with non-existent tokens return 400 error codes
+    (2) Requests with existent tokens return all matching requests + all untokenized requests after the session was created but
```
Kyle-Verhoog, line 10: I think we still need a test for the latter part of (2)

tabgok: that is tested under test_session_association_of_untokenized_traces

tests/test_session.py
```diff
@@ -132,3 +131,51 @@ async def test_session_requests(
...
+    assert len(await resp_a.json()) == 1
+    assert await resp_b.json() == []
+
+    await do_reference_v04_http_trace()
+    resp_a = await agent.get("/test/session/traces", params={"test_session_token": "sessiona"})
+    assert resp_a.status == 200
+    resp_b = await agent.get("/test/session/traces", params={"test_session_token": "sessionb"})
+    assert resp_b.status == 200
+    assert len(await resp_a.json()) == 1
+    assert len(await resp_b.json()) == 1
```
Kyle-Verhoog, line 171: Can we extend this just a tad further to see that if we sent more traces under a _new_ `sessiona` that those traces are returned rather than from the older `sessiona` session?

Kyle-Verhoog (commented):
last one, i promise!

Kyle-Verhoog (approved):
🚢 thanks @tabgok!

PR #160, opened 2024-02-12
(fix) tests: Ignore new metric

tabgok:
A new metric, metrics._dd.tracer_kr, was added to `dd-trace-py` causing snapshot tests tofail.  This metric is now added to the list of ignored snapshot fields.

### DIFF
--- a/ddapm_test_agent/trace_snapshot.py
+++ b/ddapm_test_agent/trace_snapshot.py
@@ -30,7 +30,7 @@ from .trace import trace_id as get_trace_id
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


Kyle-Verhoog (changes requested):
Release note please 🙏 

https://github.com/DataDog/dd-apm-test-agent?tab=readme-ov-file#release-notes

Then good to go

Kyle-Verhoog (approved):
🤦‍♂️  my bad, I thought this PR was updating the default ignores for the whole test agent and not just in the tests. I'll make the change to revert the release note.

PR #163, opened 2024-02-12
(chore): add error string to handle_snapshot

tabgok:
Prior to this change debugging the handle_snapshot function was difficult, because no error message beyond "INTERNAL SERVER ERROR" was returned.  After this change, the exception's message is returned.

### DIFF
--- a/ddapm_test_agent/agent.py
+++ b/ddapm_test_agent/agent.py
@@ -28,9 +28,9 @@ from urllib.parse import urlunparse
 from aiohttp import ClientResponse
 from aiohttp import ClientSession
 from aiohttp import web
+from aiohttp.web import HTTPException
 from aiohttp.web import Request
 from aiohttp.web import middleware
-from aiohttp.web import HTTPException
 from msgpack.exceptions import ExtraData as MsgPackExtraDataException
 from multidict import CIMultiDict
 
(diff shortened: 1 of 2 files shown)


Kyle-Verhoog (commented):
Oh this is a nice usability improvement. I think we can implement this generally in a middleware rather than just for this one endpoint.

Here's an example middleware that we already have
https://github.com/DataDog/dd-apm-test-agent/blob/e0eb48daaf9e2e7c2a90297b996ed747835bfb1e/ddapm_test_agent/agent.py#L99-L107

Could you check to see if we could do that approach so we have this for all endpoints, @tabgok?

ddapm_test_agent/agent.py
```diff
@@ -1035,6 +1048,7 @@ def make_app(
     app = web.Application(
         client_max_size=int(100e6),  # 100MB - arbitrary
         middlewares=[
+            handle_exception_middleware,  #type: ignore
```
Kyle-Verhoog, line 1051:
```suggestion
            handle_exception_middleware,  # type: ignore
```

ddapm_test_agent/agent.py
```diff
@@ -30,6 +30,7 @@
 from aiohttp import web
```
Kyle-Verhoog, line 30:
```suggestion
from aiohttp import web
from aiohttp.web import HTTPException
```

ddapm_test_agent/agent.py
```diff
@@ -30,6 +30,7 @@
...
 from aiohttp.web import Request
 from aiohttp.web import middleware
+from aiohttp.web import HTTPException
```
Kyle-Verhoog, line 33:
```suggestion
```

Kyle-Verhoog (approved):
nice!

PR #182, opened 2024-07-09
Build using nix 

pawelchcki:
Add ability to configure, compile and run the test agent using only nix.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,3 @@
 .venv
+result
+*.Dockerfile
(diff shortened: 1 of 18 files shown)


majorgreys (commented):
Should I be able to import ddtrace if I start python in a nix shell?

pawelchcki:
> Should I be able to import ddtrace if I start python in a nix shell?

Nope - I only touched the devShell - TBF I think I can remove it from the PR. 

My main motivation here was to be able to build the agent in a completely standalone mode.

pawelchcki:
image

Now the test agent can also be installed on android ;)

colin-higgins (approved):

PR #184, opened 2024-08-13
feat: meta_struct support

eliottness:
## Description

This PR adds support for a specific field supported since agent version v7.31 called `meta_struct` which encapsulate another layer of msgpack values inside the main msgpack chunk.

## Motivation

Appsec has been using this field to store stacktraces for multiple quarters already. We need to be able to test this feature

## Notes to reviewer

Not overly happy about this implementation but there is not obvious way to do a first round of type checking, modify the span object, and then do a second round of type-checking. So I've simply put everything in the `verify_span` function.

### DIFF
--- a/ddapm_test_agent/trace.py
+++ b/ddapm_test_agent/trace.py
@@ -71,7 +71,7 @@ class Span(TypedDict):
     meta: NotRequired[Dict[str, str]]
     metrics: NotRequired[Dict[str, MetricType]]
     span_links: NotRequired[List[SpanLink]]
-    meta_struct: NotRequired[Dict[str, dict]]
+    meta_struct: NotRequired[Dict[str, Dict[str, Any]]]
 
 
 SpanAttr = Literal[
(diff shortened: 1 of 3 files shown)


eliottness:
It seems that the CI is broken with the following error: `/home/runner/work/_temp/b3e3d4aa-56a4-4bf2-8c9c-7f47ec8f3b93.sh: line 2: rst2html.py: command not found`

Kyle-Verhoog:
ah yes ignore that error, there's some weird dependency issue that broke in the changelog job!

ddapm_test_agent/trace.py
```diff
@@ -124,6 +126,24 @@ def verify_span(d: Any) -> Span:
...
+            decoded_meta_struct = {key: msgpack.unpackb(val_bytes) for key, val_bytes in d["meta_struct"].items()}
+            for k, val in decoded_meta_struct.items():
+                assert isinstance(
+                    val, dict
+                ), f"Expected msgpack decoded value of key 'meta_struct.{k}' to be of type: 'dict', got: {type(val)}"
+                for inner_k in val:
+                    assert isinstance(
+                        inner_k, str
+                    ), f"Expected key 'meta_struct.{k}.{inner_k}' to be of type: 'str', got: {type(inner_k)}"
+            d["meta_struct"] = decoded_meta_struct
```
Kyle-Verhoog, line 146: it's pretty unintuitive that verify_span actually mutates it. Would it be possible to do it in `_trace_decoder_flexible` instead?

Kyle-Verhoog (changes requested):
just one suggestion for where the mutation logic should be, otherwise looks good to me!

ddapm_test_agent/trace.py
eliottness: @Kyle-Verhoog  The code in `_trace_decoder_flexible` looks like it's done to unmarshal only json and not msgpack. So I don't know how I should do the same for msgpack (which is the one I am interested in). I know this is not ideal to put it in `verify_span` but I did not found a better place. I can try to create a new function and call it before `verify_span` if you want but :shrug:

Kyle-Verhoog: ah crap you're right. Yeah I think a new function that does it and added to the `decode_v0{4,5,7}` methods would be the way to go then. WDYT?

eliottness: ok let me do this then

ddapm_test_agent/trace.py
```diff
@@ -391,6 +428,7 @@ def decode_v04(content_type: str, data: bytes, suppress_errors: bool) -> v04Trac
         payload = _trace_decoder_flexible(data) if suppress_errors else json.loads(data)
     else:
         raise TypeError("Content type %r not supported" % content_type)
+    _flexible_decode_meta_struct(payload)
```
Kyle-Verhoog, line 431: do we also need to call it in `decode_v05` and `decode_v07`?

Kyle-Verhoog (approved):
just one non-blocking thought. Great work. Thank you for the tests and release note 🚢

ddapm_test_agent/trace.py
```diff
@@ -168,6 +185,26 @@
...
+    if not isinstance(value, list):
+        return
+    for maybe_trace in value:
+        if not isinstance(maybe_trace, list):
+            continue
+        for maybe_span in maybe_trace:
+            if not isinstance(maybe_span, dict):
+                continue
+            if "meta_struct" in maybe_span:
+                maybe_span["meta_struct"] = _parse_meta_struct(maybe_span["meta_struct"])
```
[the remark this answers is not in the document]
Kyle-Verhoog: how do i delete this thing

ddapm_test_agent/trace.py
Kyle-Verhoog: resolved offline, only v0.4 supports meta_struct

PR #192, opened 2024-09-06
Explicitly state write permissions on publish job

Kyle-Verhoog:
[](https://datadoghq.atlassian.net/browse/VULN-8095)


Kyle-Verhoog:
the job on `main` should confirm that the workflow still works.

PR #193, opened 2024-09-12
Add agent state header to info endpoint

VianneyRuhlmann:
Add a dummy value for the `Datadog-Agent-State` header

### DIFF
--- a/ddapm_test_agent/agent.py
+++ b/ddapm_test_agent/agent.py
@@ -679,7 +679,8 @@ class Agent:
                 "client_drop_p0s": True,
                 # Just a random selection of some peer_tags to aggregate on for testing, not exhaustive
                 "peer_tags": ["db.name", "mongodb.db", "messaging.system"],
-            }
+            },
+            headers={"Datadog-Agent-State": "03e868b3ecdd62a91423cc4c3917d0d151fb9fa486736911ab7f5a0750c63824"},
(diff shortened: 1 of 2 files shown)


Kyle-Verhoog (approved):

ddapm_test_agent/agent.py
```diff
@@ -679,7 +679,8 @@ async def handle_info(self, request: Request) -> web.Response:
                 "client_drop_p0s": True,
                 # Just a random selection of some peer_tags to aggregate on for testing, not exhaustive
                 "peer_tags": ["db.name", "mongodb.db", "messaging.system"],
-            }
+            },
+            headers={"Datadog-Agent-State": "03e868b3ecdd62a91423cc4c3917d0d151fb9fa486736911ab7f5a0750c63824"},
```
bantonsson, line 683: Could some clients be confused if this is not the real sha256 of the contents?

VianneyRuhlmann: Since the hash depends on the formatting used in the json (e.g. the agent is using [indentation](https://github.com/DataDog/datadog-agent/blob/8a80bcd1c1460ba9caa97d974568bd9d0c702f3f/pkg/trace/api/info.go#L64)), I don't think it makes much sense to try to recompute the value in the tracers.

PR #198, opened 2024-10-29
Add delay for telemetry requests

kevingosse:
Extend `--trace-request-delay` to also affect trace requests.

Also, add a `test/settings` endpoint to change the value of settings on the fly.

### DIFF
--- a/releasenotes/notes/Settings-endpoint-928ae5b6e8ddc625.yaml
+++ b/releasenotes/notes/Settings-endpoint-928ae5b6e8ddc625.yaml
@@ -1,0 +1,7 @@
+---
+features:
+  - |
+    Added a /test/settings endpoint to change the settings on the fly.
+    It takes a json encoded list of key/values, that will be applied to the app dict.
+    Example:
+    { "trace_request_delay": 5 }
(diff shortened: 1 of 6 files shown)


releasenotes/notes/Settings-endpoint-928ae5b6e8ddc625.yaml
```diff
@@ -0,0 +1,7 @@
+---
+features:
+  - |
+    Added a /test/settings endpoint to change the settings on the fly.
```
Kyle-Verhoog, line 4: little bit iffy about exposing the raw config values direct to users. This'll probably result in accidental breakages to downstream test suites that depend on config values that we might accidentally change in the test agent.

If we do wanna go this route then I'll want to ensure we have tests covering the config names we want to be stable.

Kyle-Verhoog (commented):
Just one concern about the exposure of raw setting names leading to inadvertent breaking changes. Other than that looks good!

releasenotes/notes/Settings-endpoint-928ae5b6e8ddc625.yaml
kevingosse: Done in https://github.com/DataDog/dd-apm-test-agent/pull/198/commits/faf6994bd04d4ec5ce8551545cb649454ef63d0e

Kyle-Verhoog (approved):

PR #208, opened 2025-04-16
Add SNAPSHOT_REGEX_PLACEHOLDERS

bwoebi:
E.g. depending on where snapshots run, the port may be 80 or 8080. Or the path in error.stack tags will be different.


ddapm_test_agent/trace_snapshot.py
```diff
@@ -459,40 +460,57 @@ def _ordered_span(s: Span) -> OrderedDictType[str, TopLevelSpanValue]:
...
+                            event["attributes"].pop(key[23:], None)
+            elif key.startswith("span_events."):
+                if "span_events" in span:
+                    for event in span["span_events"]:
+                        event.pop(key[12:], None)  # type: ignore
+            else:
+                span.pop(key, None)  # type: ignore
+
+        if attribute_regex_replaces:  # only walk if we actually have something to replace
+            _walk_span_attributes_with_regex_replaces(span, attribute_regex_replaces)  # type: ignore
```
brettlangdon, line 513: what is the ignore here for?

does `_walk_span_attributes_with_regex_replaces` need to take a `Span` instead?

or else, we can cast the span to a `Dict[str, Any]`

```suggestion
            _walk_span_attributes_with_regex_replaces(typing.cast(Dict[str, Any], span), attribute_regex_replaces)
```

I didn't look too deeply, but I assume this is the error?

brettlangdon (commented):
some questions/nits, but otherwise cool feature!

in Python we use some "ignores", that probably would be better with this feature.

ddapm_test_agent/trace_snapshot.py
bwoebi: Yes :-)

ddapm_test_agent/trace_snapshot.py
```diff
@@ -3,7 +3,9 @@
 import logging
 import operator
 import pprint
+from re import Pattern
 import textwrap
+import typing
```
brettlangdon, line 8:
```suggestion
import typing
```

brettlangdon, line 8:
```suggestion
```

ddapm_test_agent/trace_snapshot.py
```diff
@@ -459,40 +461,57 @@ def _ordered_span(s: Span) -> OrderedDictType[str, TopLevelSpanValue]:
...
+                            event["attributes"].pop(key[23:], None)
+            elif key.startswith("span_events."):
+                if "span_events" in span:
+                    for event in span["span_events"]:
+                        event.pop(key[12:], None)  # type: ignore
+            else:
+                span.pop(key, None)  # type: ignore
+
+        if attribute_regex_replaces:  # only walk if we actually have something to replace
+            _walk_span_attributes_with_regex_replaces(typing.cast(Dict[str, Any], span), attribute_regex_replaces)
```
brettlangdon, line 514:
```suggestion
            _walk_span_attributes_with_regex_replaces(cast(Dict[str, Any], span), attribute_regex_replaces)
```

brettlangdon (approved):
minor nit on using the already imported `cast`, otherwise lgtm

PR #213, opened 2025-06-22
feat(vcr): introduce vcr endpoint to proxy 3rd party API providers

sabrenner:
Introduces a vcr endpoint to record requests sent to 3rd party API providers. Right now, the recognized providers are

- OpenAI
- Azure OpenAI
- Deepseek

The proxy snapshot server works by forwarding requests to these providers, but saving the responses for specific requests to specific endpoints in a cassette that can be replayed at a later time. This can work for one session, but requests can also be saved in a cassettes directory and persisted in the Docker image definition to persist across future versions of the test agent. To use it, the test agent can be run as normal, and the provider base url can be specified as

```python
from openai import OpenAI

client = OpenAI(base_url="http://127.0.0.1:9126/vcr/openai")
```

To properly mount the cassettes to make sure that any new cassettes created can be shared by future releases of the test agent, mount the local installation of `dd-apm-test-agent` when running the test agent image

```bash
docker run --rm \
  -p 9126:9126 \
  -v $PWD/vcr-cassettes:/vcr-cassettes \
  ghcr.io/datadog/dd-apm-test-agent/ddapm-test-agent:latest
```

replacing `$PWD/snapshot-server-cassettes` with the path to your locally installed `dd-apm-test-agent`. Or, it can be pointed against any local cassettes directory.

Pertaining to that point, you can override what directory to read cassettes from. It can be set by the `--snapshot-server-cassettes-directory` command line option, or the `VCR_CASSETTES_DIRECTORY` when running locally, although the image comes pre-packed with some cassettes.

MLOB-3011

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,28 @@ The cassettes are matched based on the path, method, and body of the request. To
 
 Optionally specifying whatever mounted path is used for the cassettes directory. The test agent comes with a default set of cassettes for OpenAI, Azure OpenAI, and DeepSeek.
 
+#### Usage in clients
+
+To use this feature in your client, you can use the `/vcr/{provider}` endpoint to proxy requests to the provider API.
+
+```python
+from openai import OpenAI
+
+client = OpenAI(base_url="http://127.0.0.1:9126/vcr/openai")
+```
+
+#### Adding new providers
+
+To add a new provider, add a supported provider in the `PROVIDER_BASE_URLS` dictionary in `ddapm_test_agent/vcr_proxy.py`, and change your tests or use case to use the new provider in the base url:
+
+```python
+base_url = "http://127.0.0.1:9126/vcr/{new_provider}"
+```
+
+And pass in a valid API key (if needed) in the way that provider expects.
+
+To redact api keys, modify the `filter_headers` list in the `get_vcr` function in `ddapm_test_agent/vcr_proxy.py`. This can be confirmed by viewing cassettes in the `vcr-cassettes` directory (or the otherwise specified directory), and verifying that any new cassettes do not contain the api key.
+
 ## Configuration
 
 The test agent can be configured via command-line options or via environment variables.


sabrenner:
i think we were using `snapshot` as the term for this server before migrating it here but it probably has conflicting terminology with existing test session snapshots 😆 would something like `provider-proxy` be better instead? open to anything!

ddapm_test_agent/agent.py
```diff
@@ -212,6 +213,17 @@ def default_value_trace_results_summary():
     }
 
 
+def default_snapshot_server_cassettes_directory():
+    """Return the default directory for snapshot server cassettes.
```
Kyle-Verhoog, line 217: instead of hardcoding this to the docker path here can we set a default `ENV SNAPSHOT_SERVER_CASSETTES_DIRECTORY=...` in the Dockerfile?

ddapm_test_agent/agent.py
```diff
@@ -1202,6 +1215,11 @@ def make_app(
             web.get("/test/trace_check/summary", agent.get_trace_check_summary),
             web.get("/test/integrations/tested_versions", agent.handle_get_tested_integrations),
             web.post("/test/settings", agent.handle_settings),
+            web.route(
+                "*",
+                "/snapshot-server/{path:.*}",
```
Kyle-Verhoog, line 1220:
```suggestion
                "/snapshot/{path:.*}",
```

ddapm_test_agent/agent.py
```diff
@@ -1227,6 +1245,7 @@ def make_app(
     app["disable_error_responses"] = disable_error_responses
     app["snapshot_removed_attrs"] = snapshot_removed_attrs
     app["snapshot_regex_placeholders"] = snapshot_regex_placeholders
+    app["snapshot_server_cassettes_directory"] = snapshot_server_cassettes_directory
```
Kyle-Verhoog, line 1248: heheh yeah the name conflict is apparent here... Maybe let's just call it "recorder" or maybe reuse the name "vcr"?

Kyle-Verhoog (commented):
@sabrenner can we strip some of the cassettes out of this PR to simplify a bit for the first go? Maybe reduce it down to just openai to begin with?

ddapm_test_agent/agent.py
sabrenner: 🤦 yes that should work!

Dockerfile
```diff
@@ -8,6 +8,8 @@ ENV SNAPSHOT_DIR=/snapshots
 
 RUN apt update && apt install -y git curl
 
+ADD snapshot-server-cassettes ./snapshot-server-cassettes
```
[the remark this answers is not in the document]
sabrenner: i think in combination with setting the [cassettes env var](https://github.com/DataDog/dd-apm-test-agent/pull/213#discussion_r2162453182) we can def move it to a different directory - but i didn't touch `src` bc i saw we just `rm` it below (looking at the PR that did it, looks like it was for build efficiency/reducing the image size). happy to move it there if maybe you think it's OK to not remove it from `src/` but remove everything else?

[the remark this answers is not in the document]
Kyle-Verhoog: ohhh right yeah we pip install everything true. Ok yeah scratch what I said - what you have is fine lool

Kyle-Verhoog (approved):
might be good to have some docs in the readme on how to add additional snapshots for providers. Some tips and tricks for how to deal with redacting api keys, etc.

Also how to use it in the clients too!

Overall great work!

PR #234, opened 2025-08-07
Initial v1 payload support via conversion

ajgajg1134:
This is an initial support change for the /v1.0/traces endpoint. For now all we will do is accept these traces and "down convert" them to the internal v4 payload representation. A future PR will add support for fetching v1 payloads in their original form to enable writing tests specifically against the v1 representation.

I'm not a frequent python dev so any feedback is welcome!


ddapm_test_agent/trace.py
```diff
@@ -702,6 +702,354 @@ class TraceChunkV07(TypedDict):
...
+def _convert_v1_chunk(chunk: Any, string_table: List[str]) -> List[Span]:
+    if not isinstance(chunk, dict):
+        raise TypeError("Chunk must be a map.")
+
+    priority, origin, decision_maker = "", "", ""
+    trace_id, trace_id_high = 0, 0
+    meta: Dict[str, str] = {}
+    metrics: Dict[str, MetricType] = {}
+    for k, v in chunk.items():
+        if k == 1:
```
brettlangdon, line 760: constants for these magic numbers would help with legibility, but not a deal breaker from me.

ddapm_test_agent/trace.py
```diff
@@ -702,6 +702,354 @@ class TraceChunkV07(TypedDict):
...
+                raise TypeError("Trace ID must be 16 bytes, got %r." % len(v))
+            # trace_id is a 128 bit integer in a bytes array, so we need to get the last 64 bits
+            trace_id = int.from_bytes(v[8:], "big")
+            trace_id_high = int.from_bytes(v[:8], "big")
+        elif k == 7:
+            decision_maker = _get_and_add_string(string_table, v)
+        else:
+            raise TypeError("Unknown key %r in v1 trace chunk" % k)
+
+    if len(spans) == 0:
```
brettlangdon, line 788: oh fun, I _think_ there are scenarios when we send a single empty trace as a test.... but I cannot remember where that is

brettlangdon: I don't think this is an issue at all, just calling it out, I _think_ we do/used to do something like this.

brettlangdon (approved):
in general looks good

PR #235, opened 2025-08-11
otel: Add OTLP logs support via HTTP endpoint

mabdinur:
## Motivation

The dd-apm-test-agent currently supports testing APM traces and metrics but lacks support for OpenTelemetry logs. As OpenTelemetry adoption grows, there's a need to test OTLP logs exporters and libraries that send logs in the standard OTLP v1.7.0 protobuf format. This gap makes it difficult to validate OpenTelemetry logs instrumentation and exporter behavior in testing environments.

## Description

This PR adds OTLP logs support to the dd-apm-test-agent by implementing a dual-server architecture that maintains separation between APM and OTLP endpoints while sharing session management.

Key Changes:
- New OTLP HTTP Server: Dedicated server on port 4318 with /v1/logs endpoint that accepts OTLP protobuf payloads
- Shared Agent Instance: Both APM (port 8126) and OTLP (port 4318) servers use the same Agent instance for unified session management
- Session Integration: OTLP logs are stored and retrievable via /test/session/logs endpoint, consistent with existing trace/metric patterns
- Future-Proof Architecture: Scaffolding for GRPC server on port 4317 that will forward to HTTP server, reusing all processing logic

Implementation Details:
- ddapm_test_agent/logs.py - OTLP protobuf decoding utilities
- make_otlp_app() - Separate aiohttp application for OTLP endpoints
- handle_v1_logs() - Processes and stores OTLP logs requests
- Comprehensive test suite with HTTP-specific naming and GRPC placeholders
- Configuration via --otlp-port and --otlp-grpc-port arguments OR OTLP_HTTP_PORT and OTLP_GRPC_PORT env vars.

Next Steps:
- Add support for grpc export and otlp metrics.

### DIFF
--- a/ddapm_test_agent/__init__.py
+++ b/ddapm_test_agent/__init__.py
@@ -1,4 +1,4 @@
 def _get_version():
-    import pkg_resources  # type: ignore[import-untyped]
+    import pkg_resources
 
     return pkg_resources.get_distribution(__name__).version
--- a/ddapm_test_agent/__init__.py
+++ b/ddapm_test_agent/__init__.py
@@ -1,4 +1,4 @@
 def _get_version():
-    import pkg_resources
+    import pkg_resources  # type: ignore[import-untyped]
 
     return pkg_resources.get_distribution(__name__).version
--- a/tests/test_logs.py
+++ b/tests/test_logs.py
@@ -1,45 +1,50 @@
 import json
+
+from opentelemetry.proto.collector.logs.v1.logs_service_pb2 import ExportLogsServiceRequest
+from opentelemetry.proto.common.v1.common_pb2 import AnyValue
+from opentelemetry.proto.common.v1.common_pb2 import KeyValue
+from opentelemetry.proto.logs.v1.logs_pb2 import LogRecord
+from opentelemetry.proto.logs.v1.logs_pb2 import ResourceLogs
+from opentelemetry.proto.logs.v1.logs_pb2 import ScopeLogs
+from opentelemetry.proto.resource.v1.resource_pb2 import Resource
 import pytest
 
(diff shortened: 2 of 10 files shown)


ddapm_test_agent/__init__.py
```diff
@@ -1,4 +1,4 @@
 def _get_version():
-    import pkg_resources  # type: ignore[import-untyped]
+    import pkg_resources
```
mabdinur, line 2:
```suggestion
    import pkg_resources  # type: ignore[import-untyped]
```

tests/test_logs.py
```diff
@@ -0,0 +1,372 @@
...
+    return f"{parsed_url.scheme}://{parsed_url.hostname}:{OTLP_HTTP_PORT}"
+
+
+@pytest.fixture
+def otlp_http_client(otlp_http_url):
+    """Test Agent client for retrieving logs from the OTLP HTTP endpoint."""
+    return TestAgentClient(otlp_http_url)
+
+
+async def test_logs_endpoint_basic_http(otlp_http_agent, otlp_logs):
```
zacharycmontoya, line 206: For these test cases, can you parametrize the following inputs and then at the beginning of the test method you could call `otlp_logs` to generate the request data from these inputs?

- service_name
- environment
- version
- log_message (this could even be an array to test sending multiple log messages from the same scope in the same request)

zacharycmontoya: Basically, I don't think the inputs should be test fixtures since we could theoretically test a variety of combinations of them

mabdinur: Oh we can parameterize fixtures. I provided an example on another test. This allows us to create default values for otlp_logs that can be overriden in other tests.

zacharycmontoya (approved):

PR #242, opened 2025-08-27
feat: add prototype test agent ui

brettlangdon:
This change adds a prototype and start to a UI for the test agent.

This adds:

### Dashboard
Showing basic config information

image

### Requests view

Shows all requests received by the trace agent in realtime, allows viewing the request/response headers and bodies, will try to decode and pretty print the bodies, and will try to give a rough waterfall view if it is a trace payload.

You can use a basic filter query language to filter requests.

The UI will keep all streamed requests until the page is refreshed. The server will only maintain the most recent 200 requests in memory.

image

image

image

### Tracer Flare

A page will the ability to trigger a tracer flare.

image

image

image

### Remote Config

Page showing the current remote config state, with the ability to modify it.

image

### DIFF
--- a/ddapm_test_agent/agent.py
+++ b/ddapm_test_agent/agent.py
@@ -2001,8 +2001,8 @@ def main(args: Optional[List[str]] = None) -> None:
     parser.add_argument(
         "--max-requests",
         type=int,
-        default=int(os.environ.get("MAX_REQUESTS", 1000)),
-        help="Maximum number of requests to keep in memory for the UI (default: 1000). Older requests are discarded when limit is reached.",
+        default=int(os.environ.get("MAX_REQUESTS", 200)),
+        help="Maximum number of requests to keep in memory for the UI (default: 200). Older requests are discarded when limit is reached.",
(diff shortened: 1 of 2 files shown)


ddapm_test_agent/agent.py
```diff
@@ -1948,6 +1961,18 @@ def main(args: Optional[List[str]] = None) -> None:
         default=os.environ.get("VCR_CI_MODE", False),
         help="Will change the test agent to record VCR cassettes in CI mode, throwing an error if a cassette is not found on /vcr/{provider}",
     )
+    parser.add_argument(
+        "--web-ui-port",
+        type=int,
+        default=int(os.environ.get("WEB_UI_PORT", 0)),
+        help="Port to serve the optional web UI (default: disabled). Example: --web-ui-port=8080",
+    )
+    parser.add_argument(
```
Kyle-Verhoog, line 1970: can we add docs to the README for both of these options? We should call out that the web ui one is experimental and not to be relied on

PR #246, opened 2025-09-05
chore: remove support for Python < 3.11

brettlangdon:


Kyle-Verhoog (approved):
does this break compatibility with system-tests? idk what python version they're on now but they were on 3.9 for quite a while and the test agent is imported there

brettlangdon:
> does this break compatibility with system-tests? idk what python version they're on now but they were on 3.9 for quite a while and the test agent is imported there

good question. I guess I assumed they were using the docker container and not running it directly, but something we definitely need to check before merging (or at least releasing) this

brettlangdon:
> https://github.com/datadog/system-tests?tab=readme-ov-file#minimal-requirements-end-to-end-testing looks like they've upgraded to 3.12 should all be good!

I asked Charles anyways to be 100% sure, but yeah, looks like 3.12 is required.

PR #249, opened 2025-09-18
feat(vcr): add ability to run in ci mode

sabrenner:
Lets the VCR 3rd party API proxy run in "CI mode", which will return a 404 for cassette files that are not found in the designated vcr cassette directory. This will ensure that, if enabled for CI for a given repository, that all cassettes are generated locally, and there are no unintended side effects from missing cassettes.

This flag defaults to `false`. I've also separated it from the `--snapshot-ci-mode` flag, but we could easily consolidate if we wanted with backwards compatibility. Just didn't want to make too many code changes unrelated to the vcr aspect of the testagent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,6 +196,10 @@ And pass in a valid API key (if needed) in the way that provider expects.
 
 To redact api keys, modify the `filter_headers` list in the `get_vcr` function in `ddapm_test_agent/vcr_proxy.py`. This can be confirmed by viewing cassettes in the `vcr-cassettes` directory (or the otherwise specified directory), and verifying that any new cassettes do not contain the api key.
 
+#### Running in CI
+
+To have the vcr proxy throw a 404 error if a cassette is not found in CI mode to ensure that all cassettes are generated locally and committed, set the `VCR_CI_MODE` environment variable or the `--vcr-ci-mode` flag in the cli tool to `true` (this value defaults to `false`).
+
 ## Configuration
 
 The test agent can be configured via command-line options or via environment variables.
--- a/releasenotes/notes/vcr-ci-mode-99dbd2940fff8d7a.yaml
+++ b/releasenotes/notes/vcr-ci-mode-99dbd2940fff8d7a.yaml
@@ -1,4 +1,4 @@
 ---
 features:
   - |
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -196,6 +196,10 @@ And pass in a valid API key (if needed) in the way that provider expects.
 
 To redact api keys, modify the `filter_headers` list in the `get_vcr` function in `ddapm_test_agent/vcr_proxy.py`. This can be confirmed by viewing cassettes in the `vcr-cassettes` directory (or the otherwise specified directory), and verifying that any new cassettes do not contain the api key.
 
+#### Running in CI
```
Kyle-Verhoog, line 199: should we just reuse the existing CI_MODE variable? Might make sense to not introduce another flag unless necessary so that future users don't accidentally miss it and face the same issue we did

brettlangdon (commented):
should we just use the existing `CI_MODE`? which we know is already enabled in a bunch of places/CIs.

IMHO, this PR is a fix to have VCR honor `CI_MODE`, I am not sure why someone would not want this enabled (also, I am pretty sure MLObs is the only users, so probably up to you)

sabrenner:
yeah the existing ci mode flag is specific for snapshots, ie `SNAPSHOT_CI` and `--snapshot-ci-mode`, but as noted in the description i do think it makes sense to coalesce into one `CI_MODE`/`--ci-mode` flag, and in that still honor the existing flag for snapshots as a backwards-compatible change 😄 so i can implement that

README.md
sabrenner: followed up oob - gonna keep with the separate ci mode variable for now, we can maybe coalesce the two snapshot & vcr variables down the line

releasenotes/notes/vcr-ci-mode-99dbd2940fff8d7a.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    vcr: adds a new `--vcr-ci-mode` flag and accompanying `VCR_CI_MODE` environment variable that, when set to `true` (defaults to `false`), will throw a 404 error if a cassette is not found in CI mode to ensure that all cassettes are generated locally and committed.
```
brettlangdon, line 4:
```suggestion
    vcr: adds a new ``--vcr-ci-mode`` flag and accompanying ``VCR_CI_MODE`` environment variable that, when set to ``true`` (defaults to ``false``), will throw a 404 error if a cassette is not found in CI mode to ensure that all cassettes are generated locally and committed.
```

PR #255, opened 2025-11-03
deprecation(vcr): remove `vcrpy` dependency 

sabrenner:
moves to deprecate the `vcrpy` requirement in favor of an in-house way of recording "cassettes" (i think we're stuck with this term unless we wanna break it but i think it still fits haha).

this is mainly because vcr is still troublesome for "starting" request contexts on top of each other, which when multiple requests are made in parallel can be troublesome. instead, this solution just reads from files as they're called, and do not start contexts around the requests that might interfere with each other.

included in this pr is a way to convert now "legacy" yaml cassette files from the vcr package to our new json files, deleting the old yaml ones. these files are very similar, but are different types just to denote the new format. i also considered 
- making the new cassettes also yaml
- not deleting the old yaml files and instead allowing users to read still from them

but, the idea anyhow is that in 2.0.0 of the test agent to drop it entirely (as noted by a todo on the top-level code path that translates yaml cassettes to in-house json cassettes), but this release line will serve as a way to transform them. not sure if this is the best way to go but open to suggestions!

### Testing
I verified this against `dd-trace-js` CI, where we are exclusively (except for `vertexai`) using the testagent vcr feature for serving api responses. this includes openai, anthropic, deepseek, and most importantly, aws-bedrockruntime, which proved troublesome and is the reason for some of the `base64` prefixing for responses.

otherwise, the existing tests needed very little modification. i included a test for legacy conversion that i'll remove in a major version as well.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -30,6 +30,9 @@ jobs:
       - uses: actions/setup-python@e797f83bcb11b83ae66e0230d6156d7c80228e7c # v6.0.0
         with:
           python-version: ${{ matrix.python-version }}
+      - name: Downgrade pip for riot compatibility
+        if: runner.os != 'Windows'
+        run: pip install --upgrade 'pip<24.1'
       - name: install deps
         run: pip install .[testing]
       - name: install pip-tools


.github/workflows/main.yml
```diff
@@ -30,8 +34,14 @@ jobs:
       - uses: actions/setup-python@e797f83bcb11b83ae66e0230d6156d7c80228e7c # v6.0.0
         with:
           python-version: ${{ matrix.python-version }}
+      - name: Downgrade pip for riot compatibility
```
Kyle-Verhoog:
```suggestion
```

maybe try without this too

sabrenner: this was needed, [c74847e](https://github.com/DataDog/dd-apm-test-agent/pull/255/commits/c74847ed1838c0bb2a949a6c0be88a597454f587) failed removing everything but adding just this part back in [d9187ef](https://github.com/DataDog/dd-apm-test-agent/pull/255/commits/d9187ef33965005cf09869a52f7584029a943494) made everything pass

PR #256, opened 2025-11-28
Allow control of test agent version

bric3:
By default, the version remains `test` however when the environment variable `TEST_AGENT_VERSION` is passed the agent will use this value.

```shell
$ TEST_AGENT_VERSION="1.2.3" python3.12 -m ddapm_test_agent.agent
```

```shell
curl http://127.0.0.1:8126/info
```

with docker
```shell
$ docker build -t dd-apm-test-agent:local .
$ docker run --rm -d --name test-agent -p 8126\
    -e TEST_AGENT_VERSION="1.2.3" \
    dd-apm-test-agent:local
```

```shell
$ curl -s http://localhost:8126/info
```

### DIFF
--- a/ddapm_test_agent/vcr_proxy.py
+++ b/ddapm_test_agent/vcr_proxy.py
@@ -465,7 +465,8 @@ async def proxy_request(
         )
 
     target_url = _url_path_join(provider_base_urls[provider], remaining_path)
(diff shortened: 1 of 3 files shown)


bric3:
@Kyle-Verhoog I'll merge it so progress can be made, but will come back with a test, probably next week.

PR #279, opened 2026-02-20
feat: validate and sort meta.events JSON encoded values

brettlangdon:
1. Add a check to ensure the value is a valid JSON array
2. Add sorting to the payload parsing and snapshot generation to help any tracers whose hashmap key encoding is not sorted/deterministic


brettlangdon:
> release note and then lgtm

@Kyle-Verhoog done, release note added

PR #325, opened 2026-04-14
Add support to lapdog for the pi coding agent

tomshen:
[Traces](https://app.datadoghq.com/llm/traces?query=%40ml_app%3Alapdog%20%40event_type%3Aspan%20%40parent_id%3Aundefined%20user_handle%3Atom.shen%40datadoghq.com&agg_m=count&agg_m_source=base&agg_t=count&fromUser=true&is_llm_session=true&is_llm_sessions_list=true&refresh_mode=paused&selectedTab=overview&sp=%5B%7B%22mp%22%3A%7B%22className%22%3A%22llm-obs-panel_WDDTtW_isSessionExpanded%22%7D%2C%22p%22%3A%7B%22eventId%22%3A%22AwAAAZ2NFxb8Aza3GAAAABhBWjJORnhiOEFBQTY3X2RIc3ozOEFBQUEAAAAkZjE5ZDhkMTctYzhmYS00MDlhLWExMTctNWRjZGY5ZjZhZWQzAAClug%22%7D%2C%22i%22%3A%22llm-obs-panel%22%7D%5D&spanId=1262887954242566239&start=1776185630721&end=1776189230721&paused=true)

```
npm install -g @mariozechner/pi-coding-agent
```

```
lapdog pi
```

### DIFF
--- a/ddapm_test_agent/claude_hooks.py
+++ b/ddapm_test_agent/claude_hooks.py
@@ -146,7 +146,11 @@ class SessionState:
         self.agent_span_stack: List[Dict[str, Any]] = []
         self.pending_tools: Dict[str, PendingToolSpan] = {}
         self.user_prompts: List[str] = []
+        # Currently active model ID (updated on each model_select / agent_start).
         self.model: str = ""
+        # Ordered history of distinct models used during this session (deduplicated
+        # on append so consecutive duplicate entries are collapsed, but the
+        # sequence is preserved — a Set would lose ordering).
         self.models: List[str] = []
         self.model_provider: str = ""
         self.tools_used: Set[str] = set()
--- a/ddapm_test_agent/claude_hooks.py
+++ b/ddapm_test_agent/claude_hooks.py
@@ -32,7 +32,6 @@ from .claude_cost_tracker import COST_METRIC_KEYS
 from .claude_link_tracker import ClaudeLinkTracker
 from .llmobs_event_platform import with_cors
 
-
 log = logging.getLogger(__name__)
 
 _HOSTNAME = socket.gethostname()
@@ -148,6 +147,8 @@ class SessionState:
         self.pending_tools: Dict[str, PendingToolSpan] = {}
         self.user_prompts: List[str] = []
         self.model: str = ""
+        self.models: List[str] = []
+        self.model_provider: str = ""
         self.tools_used: Set[str] = set()
         self.root_span_emitted: bool = False
         # Deferred agent spans waiting for PostToolUse(Task) to provide their output.
@@ -341,9 +342,7 @@ class ClaudeHooksAPI:
         """Merge key-value pairs into span['meta']['metadata']['_dd'], preserving existing values."""
         span["meta"].setdefault("metadata", {}).setdefault("_dd", {}).update(kwargs)
 
-    def _set_permission_wait_critical_evaluation(
-        self, span: Dict[str, Any], estimated_permission_wait_ms: int
-    ) -> None:
+    def _set_permission_wait_critical_evaluation(self, span: Dict[str, Any], estimated_permission_wait_ms: int) -> None:
         """Embed a permission_wait_critical boolean evaluation on a span.
         The evaluation flags whether the permission wait was > 50% of span duration.
         """
@@ -358,8 +357,7 @@ class ClaudeHooksAPI:
             "assessment": assessment,
             "status": "OK",
             "reasoning": (
-                f"Permission wait {'exceeded' if is_critical else 'did not exceed'} "
-                f"50% of span duration"
+                f"Permission wait {'exceeded' if is_critical else 'did not exceed'} " f"50% of span duration"
             ),
         }
         # Also populate legacy fields so the frontend can read them
@@ -914,9 +912,7 @@ class ClaudeHooksAPI:
             )
             entry["call_count"] += 1
             entry["total_duration_ns"] += span.get("duration", 0)
-            perm_wait = (
-                span.get("meta", {}).get("metadata", {}).get("_dd", {}).get("estimated_permission_wait_ms", 0)
-            )
+            perm_wait = span.get("meta", {}).get("metadata", {}).get("_dd", {}).get("estimated_permission_wait_ms", 0)
             if perm_wait:
                 entry["permission_wait_count"] += 1
                 entry["permission_wait_ms"] += perm_wait
@@ -959,11 +955,10 @@ class ClaudeHooksAPI:
         )
         first_breakdown = (
             first_span.get("meta", {}).get("metadata", {}).get("_dd", {}).get("context_breakdown")
-            if first_span else None
+            if first_span
+            else None
         )
-        last_breakdown = (
-            last_span.get("meta", {}).get("metadata", {}).get("_dd", {}).get("context_breakdown")
-        )
+        last_breakdown = last_span.get("meta", {}).get("metadata", {}).get("_dd", {}).get("context_breakdown")
         context_delta: Dict[str, Any] = {
             "first_input_tokens": first_input_tokens,
             "last_input_tokens": last_input_tokens,
@@ -1353,9 +1348,7 @@ class ClaudeHooksAPI:
         log.info("Resolved backend target: %s (mode=%s)", url, "agent" if agent_url else "agentless")
         return url, headers
 
-    async def _post_to_backend(
-        self, url: str, headers: Dict[str, str], data: bytes, description: str
-    ) -> None:
+    async def _post_to_backend(self, url: str, headers: Dict[str, str], data: bytes, description: str) -> None:
         """POST the given data to the url and log the outcome."""
         try:
             async with ClientSession() as http_session:
@@ -1403,9 +1396,11 @@ class ClaudeHooksAPI:
 
         # Strip locally-computed cost estimates before forwarding — let real cost tracking happen on ingestion
         forwarded_spans = [
-            {**s, "metrics": {k: v for k, v in s["metrics"].items() if k not in COST_METRIC_KEYS}}
-            if s.get("metrics")
-            else s
+            (
+                {**s, "metrics": {k: v for k, v in s["metrics"].items() if k not in COST_METRIC_KEYS}}
+                if s.get("metrics")
+                else s
+            )
             for s in spans
         ]
 
--- a/ddapm_test_agent/lapdog_cli.py
+++ b/ddapm_test_agent/lapdog_cli.py
@@ -1,4 +1,4 @@
-"""CLI for lapdog subcommands: run, stop, status, claude."""
+"""CLI for lapdog subcommands"""
 import argparse
 import os
 import shutil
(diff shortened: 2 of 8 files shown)


ddapm_test_agent/lapdog_cli.py
```diff
@@ -232,11 +265,116 @@ def cmd_claude(sub_cmd_args: List[str], forward_data: bool) -> None:
...
+
+    with open(_PI_EXT_SOURCE, "r") as f:
+        source = f.read()
+
+    # Patch the default URL if using a non-standard port.
+    lapdog_url = f"http://localhost:{port}"
+    source = source.replace(
+        'const LAPDOG_URL = process.env.LAPDOG_URL || "http://localhost:8126";',
+        f'const LAPDOG_URL = process.env.LAPDOG_URL || "{lapdog_url}";',
+    )
```
sabrenner: i think if port is non-standard, we should just pass in an updated `LAPDOG_URL` when starting the `pi` process, if it passes process environment variables down into the extensions. ideally we don't modify the source directly, in case we change it in the future.

ddapm_test_agent/lapdog_cli.py
```diff
@@ -232,11 +265,116 @@ def cmd_claude(sub_cmd_args: List[str], forward_data: bool) -> None:
...
+                existing = f.read()
+            if existing == source:
+                print(f"[lapdog] pi extension already installed at {_PI_EXT_DEST}")
+                return
+            is_update = True
+        except OSError:
+            pass
+
+    with open(_PI_EXT_DEST, "w") as f:
+        f.write(source)
```
sabrenner, line 310: just to cut down on os-level operations, should we only do this if `is_update` is `True`?

ddapm_test_agent/lapdog_cli.py
```diff
@@ -232,11 +265,116 @@ def cmd_claude(sub_cmd_args: List[str], forward_data: bool) -> None:
...
+
+def _uninstall_pi_extension() -> None:
+    """Remove the lapdog extension if it was installed by us."""
+    if not os.path.isfile(_PI_EXT_DEST):
+        return
+    try:
+        with open(_PI_EXT_DEST, "r") as f:
+            content = f.read()
+        if _PI_EXT_MARKER not in content and "lapdog" not in content[:200]:
+            # Not our file — leave it alone.
```
sabrenner: feels a little janky but is probably safe enough 😆

ddapm_test_agent/lapdog_cli.py
```diff
@@ -232,11 +265,116 @@ def cmd_claude(sub_cmd_args: List[str], forward_data: bool) -> None:
...
+
+def _run_pi(args: Optional[List[str]] = None) -> None:
+    """Exec the pi binary, forwarding arguments.  Never returns."""
+    if args is None:
+        args = []
+    pi_bin = shutil.which("pi")
+    if not pi_bin:
+        print("[lapdog] 'pi' not found in PATH", file=sys.stderr)
+        sys.exit(1)
+    os.execv(pi_bin, [pi_bin] + args)
```
sabrenner: i think at this point we can set the `LAPDOG_URL` arg with the `port` (will need to be passed into this function)

ddapm_test_agent/lapdog_cli.py
```diff
@@ -232,11 +265,116 @@ def cmd_claude(sub_cmd_args: List[str], forward_data: bool) -> None:
...
+    with open(_PI_EXT_DEST, "w") as f:
+        f.write(source)
+
+    if is_update:
+        print(f"[lapdog] Updated pi extension → {_PI_EXT_DEST}")
+    else:
+        print(f"[lapdog] Installed pi extension → {_PI_EXT_DEST}")
+
+
+def _uninstall_pi_extension() -> None:
```
sabrenner: also i saw that we're not surfacing this anywhere, we might be able to just remove it

ddapm_test_agent/pi_hooks.py
```diff
@@ -0,0 +1,547 @@
...
+        model_provider = body.get("model_provider", "")
+        if model_id:
+            session.model = model_id
+        log.info("Pi session started: %s (model=%s/%s)", session_id, model_provider, model_id)
+
+    def _handle_model_select(self, session_id: str, body: Dict[str, Any]) -> None:
+        session = self._get_or_create_session(session_id)
+        model_id = body.get("model_id", "")
+        if model_id:
+            session.model = model_id
```
sabrenner, line 105: question - do we want to record all model id's used during a session? or are we fine with overwriting it every time.

ddapm_test_agent/pi_lapdog_extension.ts
```diff
@@ -0,0 +1,247 @@
...
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+function truncate(value: unknown, maxBytes: number = MAX_OUTPUT_BYTES): string {
+	const str = typeof value === "string" ? value : JSON.stringify(value ?? "");
+	if (str.length <= maxBytes) return str;
+	return str.slice(0, maxBytes) + `\n... (truncated, ${str.length} bytes total)`;
+}
```
sabrenner: do we wanna truncate stuff at this level?

ddapm_test_agent/pi_lapdog_extension.ts
sabrenner, line 1: just a thought that we could make this logic even simpler by removing any sort of processing and just forward the whole event to our hooks and let it do the processing instead

sabrenner (approved):
lgtm! just left some cleanup thoughts but not blocking as it looks like it works locally for you, and we can iterate more if needed

ddapm_test_agent/pi_hooks.py
tomshen: hmm we should probably capture all of them: you can switch models mid-session pretty easily (and I think the resulting instrumentation gets a little messy)

sabrenner: ok cool, i think it's maybe fine for now since we'll aggregate the models in the span view anyway, but if we wanted to tag the top-level agent span with all the models it used as well, we'd just have to change up `SessionState` slightly to record a list of models instead of a single model.

ddapm_test_agent/lapdog_cli.py
tomshen: Done — I stopped patching the installed TypeScript source entirely. `_install_pi_extension()` now just copies the bundled extension as-is, and the resolved lapdog URL is passed at runtime via `LAPDOG_URL` when launching `pi`.

ddapm_test_agent/lapdog_cli.py
tomshen: Yep — wired that through now. `_run_pi()` takes the resolved port and uses `os.execve(...)` with `LAPDOG_URL=http://localhost:{port}` in the child environment.

ddapm_test_agent/lapdog_cli.py
tomshen: Updated this — `os.makedirs()` now only happens right before we write the file, after the up-to-date early return, so we skip that extra filesystem work on no-op installs.

ddapm_test_agent/lapdog_cli.py
tomshen: Removed it. `_PI_EXT_MARKER` is gone now, and uninstall just checks the installed file header for `lapdog` before deleting it.

ddapm_test_agent/pi_lapdog_extension.ts
tomshen: Went with this simplification. The extension is now a thin forwarder for raw `agent_end` / `message_end` / tool payloads, and all extraction/parsing lives in `pi_hooks.py`.

ddapm_test_agent/pi_hooks.py
tomshen: Updated this to track the full model history + provider changes. `SessionState` now keeps `models` and `model_provider`; `session_start` / `model_select` append model changes and update provider; the root agent span stores `models_used`, and LLM/tool spans are stamped with the current model + provider at emit time.

ddapm_test_agent/pi_lapdog_extension.ts
tomshen: I moved truncation out of the extension. The extension now forwards the raw event payloads, and `pi_hooks.py` does the truncation while assembling spans so that processing stays in one place.

ddapm_test_agent/claude_hooks.py
```diff
@@ -148,6 +147,8 @@ def __init__(self, session_id: str, trace_id: str, root_span_id: str, start_ns:
         self.pending_tools: Dict[str, PendingToolSpan] = {}
         self.user_prompts: List[str] = []
         self.model: str = ""
+        self.models: List[str] = []
+        self.model_provider: str = ""
```
sabrenner, line 155: i would maybe just add a small comment on the distinction between `model` and `models`. we could also make `models` a `Set`

ddapm_test_agent/lapdog_cli.py
sabrenner: sorry, i meant the function itself isn't being used anywhere 😅 so we can remove this function.

ddapm_test_agent/pi_lapdog_extension.ts
sabrenner: file looks better without the truncation logic! fwiw - in my orig comment i had meant each event listener could look something like
```ts
pi.on("{event}", (event, ctx) => post("{event}", { event, ctx })
```
such that its even more generic. however, i do see we do stuff like `ctx.sessionManager.getSessionId()`, which would not translate well to the above model, i think it's fine to leave as-is

ddapm_test_agent/pi_hooks.py
```diff
@@ -0,0 +1,613 @@
...
+from .claude_hooks import _format_trace_id
+from .claude_hooks import _to_json_str
+from .llmobs_event_platform import with_cors
+
+log = logging.getLogger(__name__)
+
+_MAX_CONTENT_BYTES = 128 * 1024  # 128 KB cap — mirrors the limit previously in the extension
+
+
+def _truncate(value: str, max_bytes: int = _MAX_CONTENT_BYTES) -> str:
```
sabrenner: i'm still a little confused why we truncate at all

sabrenner (approved):
nice, lgtm! we can merge, just left some follow up thoughts that idt are too worth applying here if it works as-is

ddapm_test_agent/claude_hooks.py
tomshen: Added comments clarifying the distinction. Kept it as a `List` rather than `Set` since order matters (we want to know the sequence of model switches) and we already deduplicate on append.

ddapm_test_agent/lapdog_cli.py
tomshen: Removed!

ddapm_test_agent/pi_hooks.py
tomshen: Good call — truncation is handled downstream by EvP so it's redundant here. Removed `_truncate()` and all call sites.

releasenotes/notes/add-lapdog-pi-command-for-tracing-pi-coding-agent-38926d3a88dfe33d.yaml
```diff
@@ -0,0 +1,7 @@
+---
+features:
+  - |
+    Add `lapdog pi` command for tracing pi coding agent sessions through the
```
sabrenner, line 4:
```suggestion
    Add ``lapdog pi`` command for tracing pi coding agent sessions through the
```

for the future, i missed it in this PR xD

PR #328, opened 2026-04-15
Emit step span kind from pi hooks

tomshen:
## What

Introduce synthetic `step` spans in pi hooks that group each inference cycle (one LLM call + its downstream tool calls) within a pi agent turn.

This follows the [trajectory-dev proposal](https://github.com/DataDog/trajectory-dev/blob/main/docs/proposals/LLMOBS-TURN-SPAN-KIND.md) (LLMOBS-TURN-SPAN-KIND.md).

## Trace shape

```
agent(turn)                          ← agent_start / agent_end
  step                               ← turn_start / turn_end
    llm                              ← message_start / message_end
    tool
    tool
  step
    llm
    tool
  step                               ← final response, no tools
    llm
```

### `ddapm_test_agent/pi_hooks.py`
- New `ActiveStepSpan` class tracks the active inference cycle per session
- New `_extract_output_text_and_tool_calls()` parses assistant content blocks
- `_start_step()` / `_finalize_active_step()` manage step span lifecycle
- `turn_start` opens a step span (primary); `message_start` opens a fallback step if none exists
- `turn_end` finalizes the step (primary); `agent_end` / `session_shutdown` catch unflushed steps
- LLM spans (`message_end`) and tool spans parent under the active step
- Root agent span tagged `trajectory.semantic_type:turn`
- Step spans tagged `trajectory.semantic_type:agent_message`
- Step metadata includes `message_index`, `turn_index`, `tool_use_ids`, `has_thinking`, `stop_reason`

### `tests/test_pi_hooks.py` (new)
12 tests covering:
- Basic parenting: `agent → step → llm`
- Tools parent under step
- Multiple sibling steps
- Trajectory tags
- LLMObs list API filtering by `@meta.span.kind:step`
- Finalization on `agent_end` (no `turn_end`), `session_shutdown`
- Fallback step when `message_start` arrives without `turn_start`
- Step index reset across agent cycles
- Thinking block detection
- Cancelled turn (no message)

## Terminology mapping

| Pi event | Proposal concept | Span kind |
|----------|-----------------|-----------|
| `agent_start` / `agent_end` | **turn** | `agent` |
| `turn_start` / `turn_end` | **step** | `step` |
| `message_start` / `message_end` | LLM call | `llm` |

### DIFF
--- a/ddapm_test_agent/pi_hooks.py
+++ b/ddapm_test_agent/pi_hooks.py
@@ -7,16 +7,22 @@ span storage so that pi traces appear alongside Claude Code traces in the UI.
 Pi extension events and their mapping:
 
     session_start     → create session, set model
-    agent_start       → start new trace (like UserPromptSubmit)
-    agent_end         → finalize root span (like Stop)
-    message_start     → begin tracking an LLM span
+    agent_start       → start new trace / root agent span (proposal "turn")
+    agent_end         → finalize root span
+    turn_start        → open a step span (proposal "step" = one inference cycle)
+    turn_end          → finalize step span
+    message_start     → begin tracking an LLM span (child of step)
     message_end       → emit LLM span with token usage
-    tool_execution_start → create pending tool span (like PreToolUse)
-    tool_execution_end   → emit tool span (like PostToolUse)
-    turn_start / turn_end → logged, no span emitted
+    tool_execution_start → create pending tool span (child of step)
+    tool_execution_end   → emit tool span
     model_select      → update session model
     session_compact   → mark compaction on current span
     session_shutdown  → finalize session (like SessionEnd)
+
+Terminology mapping (pi → trajectory-dev proposal):
+    pi agent_start/agent_end  →  proposal "turn" (full response to one user input)
+    pi turn_start/turn_end    →  proposal "step" (one inference cycle + its tools)
+    pi message_start/end      →  proposal LLM call within a step
 """
 
 import json
@@ -26,6 +32,7 @@ from typing import Any
 from typing import Dict
 from typing import List
 from typing import Optional
+from typing import Tuple
 
 from aiohttp import web
 from aiohttp.web import Request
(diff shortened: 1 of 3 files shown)


ddapm_test_agent/pi_hooks.py
```diff
@@ -81,6 +150,117 @@ def _current_parent_id(self, session: SessionState) -> str:
...
+
+    # ------------------------------------------------------------------
+    # Step lifecycle helpers
+    # ------------------------------------------------------------------
+
+    def _start_step(self, session: SessionState, turn_index: Optional[int] = None) -> ActiveStepSpan:
+        """Open a new step span.  Finalizes any prior active step first."""
+        sid = session.session_id
+        # Finalize prior step if still open
+        self._finalize_active_step(sid)
```
sabrenner, line 168: is it ever possible for multiple pi steps to be running at once? i imagine not but just wanted to double check as otherwise this could break

sabrenner (approved):
one question but overall lgtm!

ddapm_test_agent/pi_hooks.py
tomshen: Good question — no, pi processes one inference cycle (step) at a time per session. It's single-threaded: turn_start → LLM call → tool executions → turn_end, then the next step begins. No parallelism within a session.

Additionally, `_start_step` defensively calls `_finalize_active_step(sid)` first (line 168), so even if events arrived out of order the prior step would get finalized before a new one opens.

tomshen: Actually, small correction to my earlier reply — "single-threaded" was sloppy. Pi does support **parallel tool execution** (it's the default mode), so multiple `tool_execution_start`/`tool_execution_end` events can interleave within a single step. But that doesn't affect the step tracking here because all parallel tools belong to the same active step — they all get the same parent via `_active_step_parent_id()`.

Steps themselves are strictly sequential: the agent loop in pi-mono's `agent-loop.ts` is a `while` loop that does `turn_start` → LLM call → tool executions → `turn_end` → next turn. There's no sub-agent spawning or concurrent step execution. So only one step is ever active per session at a time.

PR #329, opened 2026-04-20
Implement /api/ui/query/scalar for LLMObs data sources

tomshen:
## Summary

Replaces the stub `POST /api/ui/query/scalar` handler with a real implementation backed by the in-memory LLMObs spans tracked by the test agent. This unblocks lapdog local-dev tiles (Application Overview, AI Agents Console, trace detail panels) that currently get empty data back.

## What's supported

- **Data sources**: `llm_observability` and `llm_observability_stream`. Unknown data sources return a zero-valued number column instead of a 500 so tiles render "no data".
- **Aggregations**: `count`, `cardinality`, `sum`, `avg`, `min`, `max`, `median`, `pc75`/`pc90`/`pc95`/`pc99`, `latest`, `earliest`.
- **group_by** with per-dimension sort order + row limit.
- **search.query** via the existing `apply_filters` / `parse_filter_query` helpers (same syntax used by `/aggregate` and `/list`).
- **Per-request time windows** (`from`/`to` in ms).
- **Formulas** as direct query-name references (e.g. `{formula: 'q1', alias: 'span_count'}`). Arithmetic formulas are deferred.

## Tests

8 new tests in `tests/test_llmobs_event_platform.py` covering: count w/o group_by, group_by kind, avg(@duration), time-window filtering, search.query filtering, unknown data source fallback, multi-request bodies, and formula references.

Legacy empty-input behaviour (`{data: []}` → one empty `scalar_response`) is preserved so the existing stub tests still pass. All 38 tests in the file pass; flake8 and mypy are clean for the touched files.

## Out of scope / follow-ups

- Formula arithmetic (`query1 / query2`, literals, parens) — punt until a caller actually needs it.
- Timeseries endpoint (`/api/ui/query/timeseries`) — separate endpoint, separate PR.
- Per-session scoping via `test_session_token` — matches current `/aggregate` behaviour; add together if/when we retrofit both.

### DIFF
--- a/ddapm_test_agent/llmobs_event_platform.py
+++ b/ddapm_test_agent/llmobs_event_platform.py
@@ -6,6 +6,7 @@ import gzip
 import json
 import logging
 import re
+import statistics
 import time
 from typing import Any
 from typing import Awaitable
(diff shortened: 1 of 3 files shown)


sabrenner (approved):
RSLGTM, tested this out against the related frontend changes and all seems good, and didn't see any anti-patterns in tests! 🚢

PR #334, opened 2026-04-21
feat(claude-hooks): add step span kind for inference cycles

tomshen:
## Summary

Insert a `step` span between the agent span and its `llm`/`tool` children for instrumented Claude Code sessions, so each **inference cycle** (one LLM call + the tools it dispatches) becomes a first-class unit in the trace tree. Follows the [LLMOBS-TURN-SPAN-KIND proposal](https://github.com/DataDog/trajectory-dev/blob/main/docs/proposals/LLMOBS-TURN-SPAN-KIND.md) and mirrors the existing implementation for the PI agent in `pi_hooks.py`.

### Before (flat, no step layer)

```
agent (turn)
├── llm
├── tool (Read)
├── tool (Bash)
├── llm
├── tool (Edit)
└── llm
```

### After (step per inference cycle)

```
agent (turn)
├── step (inference-0)
│   ├── llm
│   ├── tool (Read)
│   └── tool (Bash)
├── step (inference-1)
│   ├── llm
│   └── tool (Edit)
└── step (inference-2)    ← final response, no tools
    └── llm
```

Non-instrumented sessions (no LLM proxy) keep the existing flat hierarchy — steps only exist when LLM boundaries are observable.

## Design

- **Driving signal**: `claude_proxy._create_llm_span`. Each LLM call on an instrumented session opens one step.
- **Tool-to-step parenting**: reuses the existing `claude_link_tracker._llm_span_parents` map. `_start_step_for_llm` updates the map to point at the step span_id, so `get_parent_for_tool(tool_use_id)` transparently resolves tools to the step — no changes in `_handle_pre_tool_use` / `_handle_post_tool_use`.
- **Agent-owner hint is separate from tree parent**: `TrackedToolCall` carries both `tool_parent_id` (the tree parent, a step) and `tool_agent_id` (the agent that owns the tool). `on_tool_call_output_used` returns the agent id as the parent hint, so consecutive LLM calls resolve to the agent — steps open as siblings, never nested inside each other. `_start_step_for_llm` also dereferences any step span id it receives as `agent_parent_id` (via `SessionState.step_agent_by_span_id`) as a defensive climb.
- **Per-agent active step**: `SessionState.active_steps_by_agent` is keyed by agent span_id so nested subagents each carry an independent step.
- **Eager emission**: step span is written with `duration=0` and updated in-place on finalize — matches the existing pattern for root / subagent spans.
- **Finalization triggers**: next LLM call on the same agent (ends at new LLM's `start_ns`, no overlap), `SubagentStop`, `Stop`, interrupted turn, trace reset on new `UserPromptSubmit`.
- **Step metadata**: `meta.span.kind = "step"`, `name = "inference-{N}"`, tag `trajectory.semantic_type:agent_message`, `meta.metadata.{message_index, tool_use_ids, has_thinking, stop_reason}`, `meta.output.value` = concatenated assistant text blocks.
- **Parent re-resolution at PostToolUse**: `PreToolUse` can fire before the streaming proxy finishes `_create_llm_span`, in which case `pending.parent_id` falls back to the current agent rather than the step. `_handle_post_tool_use`, `_handle_post_tool_use_failure`, and `_handle_subagent_start` re-query the link tracker via `_resolve_tool_parent` so any stale agent parent is upgraded to the step once the LLM span is registered.

## Files changed

| File | Purpose |
|---|---|
| `ddapm_test_agent/claude_hooks.py` | `ActiveStep` class, `SessionState` fields (`active_steps_by_agent`, `step_index_by_agent`, `step_agent_by_span_id`), `_agent_for_span_id` helper, step finalization wiring. |
| `ddapm_test_agent/claude_proxy.py` | Six-line integration in `_create_llm_span`. |
| `ddapm_test_agent/claude_link_tracker.py` | `TrackedToolCall.tool_agent_id`; `on_tool_call` takes the agent id; `on_tool_call_output_used` returns it as the parent hint. |
| `tests/test_claude_steps.py` | **New**, 14 test cases. |
| `tests/test_claude_hooks.py` | Regression assertion: non-instrumented sessions produce no step spans. |
| `releasenotes/notes/claude-hooks-step-span-kind-*.yaml` | Release note. |

## Test plan

- [x] `pytest tests/test_claude_steps.py` — 14/14 passing
  - instrumented session creates step; non-instrumented does not
  - multiple inferences produce sibling steps (`inference-0/1/2`)
  - step without tools (text-only response)
  - step duration chains exactly (`step[n].start + step[n].duration == step[n+1].start`)
  - subagent has its own step, parented to the subagent span
  - interrupted turn finalizes open step with `status="error"` and `meta.error.message="interrupted"`
  - stop finalizes open step with populated metadata
  - step metadata: `message_index`, `tool_use_ids`, `has_thinking`, `stop_reason`
  - step output value = concatenated assistant text blocks
  - step carries `trajectory.semantic_type:agent_message` tag
  - concurrent subagents each have their own step, no cross-parenting
  - **consecutive inferences with tool_result remain siblings** (regression: previously step-(N+1) parented to step-N because the tool_result parent hint returned the tool's tree parent, a step)
  - **PreToolUse/LLM-span-creation race** (regression: parallel tools from the same inference where PreToolUse fired before `_create_llm_span` completed landed under the agent instead of the step; parent is now re-resolved at PostToolUse)
- [x] `pytest tests/test_claude_hooks.py tests/test_pi_hooks.py` — 32/32 passing (regression)
- [x] `riot run -s flake8 ddapm_test_agent tests` — passes
- [x] `riot run -s mypy` — passes (54 source files, 0 issues)
- [ ] Manual end-to-end sanity check with a real Claude Code session via lapdog's LLM proxy
```sh
pip install git+https://github.com/Datadog/dd-apm-test-agent@feat/claude-hooks-step-span-kind
lapdog claude
```

Claude session: `3dc4a8f3-6e58-45dc-8bb6-e4bb9e88a02e`
Resume: `claude --resume 3dc4a8f3-6e58-45dc-8bb6-e4bb9e88a02e`


ddapm_test_agent/_clock.py
sabrenner, line 1: i personally dont know enough about this problem in python but the code here looks pretty minimal so i think it should be ok!

releasenotes/notes/time-ns-precision-27547e49e5511f28.yaml
sabrenner, line 1: we probs don't need to call this out

tests/test_claude_steps.py
```diff
@@ -0,0 +1,720 @@
...
+    if start_ns is None:
+        start_ns = _time.time_ns()
+    session = proxy_api._hooks_api._sessions.get(session_id)
+    span = proxy_api._create_llm_span(
+        session=session,
+        request_body=request_body or _request(),
+        response_data=response_data,
+        start_ns=start_ns,
+        duration_ns=duration_ns,
+    )
```
sabrenner, line 128: this feels like a bit of an anti-pattern but since we're tryna move quickly i think it's ok. just as note (probably more for myself) - we should test against the real claude sdk cli (somehow??) and just use prerecorded responses via the testagent vcr. but, that's for a later point.

i think the content these tests are testing is OK despite how we're simulating it all.

releasenotes/notes/claude-hooks-step-span-d18ee340ccf2d99a.yaml
```diff
@@ -0,0 +1,15 @@
+---
+features:
+  - |
+    claude hooks: Insert a ``step`` span between the agent span and its
```
sabrenner, line 4: i think we can just keep the first sentence as a high-level release note without explaining the specifics of our implementation

sabrenner (approved):
code looks good at a glance, just a couple comments on the release notes & one comment for myself on our testing style with this stuff but nothing actionable here. feel free to merge for now, i will try it out manually later but i know you've also been trying it out!

PR #338, opened 2026-04-28
feat(lapdog): add support for cost tracking `opus-4-7` with Claude Code

sabrenner:
Adds support for properly cost tracking `ops-4-7` via Claude Code, and addresses a bug in the local cost tracking logic.


tomshen (approved):
Code + QA LGTM

PR #339, opened 2026-04-29
feat(agent): add authentication field to info endpoint

sabrenner:
Adds an `authenticated` field to the `/info` that returns whether or not a valid `dd_api_key` and `dd_site` are present when running the agent. Has no impact on existing usage and just exposes the `_is_valid_api_key_and_site_combination` check we were already doing.

### DIFF
--- a/ddapm_test_agent/agent.py
+++ b/ddapm_test_agent/agent.py
@@ -2026,7 +2026,6 @@ def make_app(
     app["dd_site"] = dd_site
     app["dd_api_key"] = dd_api_key
 
-    # if dd_api_key and dd_site and not disable_llmobs_data_forwarding:
     valid_auth = _is_valid_api_key_and_site_combination(dd_api_key, dd_site) if dd_api_key and dd_site else False
     app["authenticated"] = valid_auth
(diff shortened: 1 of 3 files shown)


tomshen (approved):

ddapm_test_agent/agent.py
```diff
@@ -2023,11 +2026,13 @@ async def _cleanup_claude_proxy(app: web.Application) -> None:
     app["dd_site"] = dd_site
     app["dd_api_key"] = dd_api_key
 
-    if dd_api_key and dd_site and not disable_llmobs_data_forwarding:
-        valid_auth = _is_valid_api_key_and_site_combination(dd_api_key, dd_site)
-        if not valid_auth:
-            log.warning("Cannot forward LLM Observability data with an invalid DD_API_KEY and DD_SITE, disabling LLM Observability data forwarding.")
-            disable_llmobs_data_forwarding = True
+    # if dd_api_key and dd_site and not disable_llmobs_data_forwarding:
```
sabrenner:
```suggestion
```

PR #340, opened 2026-04-29
context breakdown fixes

ncybul:
This PR:
- Adds opus 4.7 to list of models with 1m token context window
- Updates the logic for finding llm spans belonging to an agent turn (this broke when we started nesting llm spans within step spans)


sabrenner (approved):
i would say a release note for this change would be good, but lgtm!

& not sure if we have test coverage for it but could be nice to have one as well

PR #346, opened 2026-05-06
chore(lapdog): add `lapdog_forwarded` tag to forwarded LLM Observability spans

sabrenner:
Adds `lapdog_forwarded:true` tags to all spans forwarded to an actual DD account when run from the `lapdog` binary.

### DIFF
--- a/ddapm_test_agent/agent.py
+++ b/ddapm_test_agent/agent.py
@@ -6,6 +6,7 @@ from collections import OrderedDict
 from collections import defaultdict
 from dataclasses import dataclass
 from dataclasses import field
+import gzip
 import json
 import logging
 import os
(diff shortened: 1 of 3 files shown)


ddapm_test_agent/agent.py
```diff
@@ -889,8 +918,12 @@ async def handle_evp_proxy_v2_api_v2_llmobs(self, request: Request) -> web.Respo
             url = f"https://llmobs-intake.{dd_site}/api/v2/llmobs"
             headers["DD-API-KEY"] = dd_api_key
 
+        raw = await request.read()
+        if request.app["lapdog_mode"]:
+            raw = _inject_lapdog_forwarded(raw, request.headers.get("Content-Encoding", ""))
```
Kyle-Verhoog, line 923: lol wait can we just do it on send vs decoding + encoding for every request?

sabrenner: i don't thinkkkkk so because these are the raw payloads we get from tracers, not the ones we create ourselves from claude/pi hooks

but i could be misunderstanding 😛

Kyle-Verhoog (approved):

Issue #13, opened 2021-09-10
Fix distributed/partial traces


Kyle-Verhoog:
Distributed traces cause the agent to err out since the parent span doesn't exist

Issue #12, opened 2021-09-09
Rename repository


Kyle-Verhoog:
The package and project name have been changed over to ddapm-test-agent, the repo should follow suit at some point.

Issue #10, opened 2021-07-29
Add ability to return the recieved traces


brettlangdon:
It would be great to be able to access the actual traces/spans sent to the test agent for additional assertions.

Being able to get the traces back and assert them would help clarify intention in test cases and make it easier for writers to assert exact what is important, and for reviewers to more easily review the expected outcomes.

PR #350, opened 2026-05-11
Add Codex tracing support to lapdog

tomshen:
## Summary
- add Lapdog Codex launcher support with a JSONL watcher that posts Codex session records to the local hook endpoint
- translate Codex JSONL records into LLMObs agent, step, LLM, and tool spans, including task completion output and preserved LLM input/output messages across tool calls
- add an OpenAI-compatible Codex proxy module for LLM span capture and make it part of the tracked patch
- align Codex trace shape with the step proposal: tool calls stay in the same step as the LLM that generated their arguments, each step has at most one LLM span, and empty `codex-model` placeholder LLM spans are not emitted
- add OpenAI cost estimates from Codex token usage and tests for list endpoint visibility, watcher behavior, CLI startup, duplicate replay, multi-LLM trace shape, proxy span capture, empty-output filtering, and late output/tool ordering
- add a release note for `lapdog codex` tracing support

## Verification
- `PYENV_VERSION=3.11.15 riot run -s flake8`
- `PYENV_VERSION=3.11.15 riot run -s fmt`
- `PYENV_VERSION=3.11.15 riot run -s mypy`
- `python -m pytest tests/test_codex_hooks.py tests/test_codex_watcher.py tests/test_lapdog_cli.py`
- `python -m pytest tests/test_codex_hooks.py tests/test_codex_proxy.py tests/test_codex_watcher.py tests/test_lapdog_cli.py`

### DIFF
--- a/ddapm_test_agent/codex_cost_tracker.py
+++ b/ddapm_test_agent/codex_cost_tracker.py
@@ -1,0 +1,76 @@
+"""Cost tracking for OpenAI models used by Codex.
+
+Pricing is in nanodollars per token (1 nanodollar = 1e-9 USD), matching the
+metric keys expected by the web-ui LLM observability span detail view.
+
+Pricing data last updated 2026-05 from OpenAI API pricing pages. GPT-5.5
+standard rates are documented for context lengths under 270K tokens.
+"""
+
+from dataclasses import dataclass
+from typing import Dict
+from typing import List
+from typing import Optional
+from typing import Tuple
+
+
+@dataclass(frozen=True)
+class _OpenAIPrice:
+    prefix: str
+    input_price: int
+    cached_input: int
+    output: int
+
+
+_PRICING: List[_OpenAIPrice] = [
+    _OpenAIPrice("gpt-5.5", input_price=5_000, cached_input=500, output=30_000),
+    _OpenAIPrice("gpt-5.4-mini", input_price=750, cached_input=75, output=4_500),
+    _OpenAIPrice("gpt-5.4", input_price=2_500, cached_input=250, output=15_000),
+    _OpenAIPrice("gpt-5.2-codex", input_price=1_750, cached_input=175, output=14_000),
+    _OpenAIPrice("gpt-5.2", input_price=1_750, cached_input=175, output=14_000),
+    _OpenAIPrice("gpt-5.1-codex-max", input_price=1_250, cached_input=125, output=10_000),
+    _OpenAIPrice("gpt-5.1-codex", input_price=1_250, cached_input=125, output=10_000),
+    _OpenAIPrice("gpt-5.1", input_price=1_250, cached_input=125, output=10_000),
+    _OpenAIPrice("gpt-5-codex", input_price=1_250, cached_input=125, output=10_000),
+    _OpenAIPrice("gpt-5", input_price=1_250, cached_input=125, output=10_000),
+    _OpenAIPrice("gpt-4.1-mini", input_price=400, cached_input=100, output=1_600),
+    _OpenAIPrice("gpt-4.1-nano", input_price=100, cached_input=25, output=400),
+    _OpenAIPrice("gpt-4.1", input_price=2_000, cached_input=500, output=8_000),
+    _OpenAIPrice("gpt-4o-mini", input_price=150, cached_input=75, output=600),
+    _OpenAIPrice("gpt-4o", input_price=2_500, cached_input=1_250, output=10_000),
+]
+
+_PREFIXES: List[Tuple[str, _OpenAIPrice]] = [(p.prefix, p) for p in _PRICING]
+
+
+def _find_price(model_id: str) -> Optional[_OpenAIPrice]:
+    normalized = model_id.lower()
+    for prefix, price in _PREFIXES:
+        if normalized == prefix or normalized.startswith(prefix + "-"):
+            return price
+    return None
+
+
+def compute_openai_cost_metrics(
+    model_id: str,
+    non_cached_input_tokens: int,
+    cached_input_tokens: int,
+    output_tokens: int,
+) -> Dict[str, int]:
+    price = _find_price(model_id)
+    if price is None:
+        return {}
+
+    non_cached_input_cost = non_cached_input_tokens * price.input_price
+    cache_read_cost = cached_input_tokens * price.cached_input
+    output_cost = output_tokens * price.output
+    input_cost = non_cached_input_cost + cache_read_cost
+
+    return {
+        "estimated_non_cached_input_cost": non_cached_input_cost,
+        "estimated_cache_write_input_cost": 0,
+        "estimated_cache_read_input_cost": cache_read_cost,
+        "estimated_input_cost": input_cost,
+        "estimated_output_cost": output_cost,
+        "estimated_total_cost": input_cost + output_cost,
+    }
--- a/ddapm_test_agent/codex_hooks.py
+++ b/ddapm_test_agent/codex_hooks.py
@@ -1,0 +1,736 @@
+"""Codex session JSONL -> LLM Observability spans."""
+
+import datetime
+import json
+import logging
+import os
+import socket
+from typing import Any
+from typing import Dict
+from typing import List
+from typing import Optional
+from typing import Set
+from typing import Tuple
+
+from aiohttp import web
+from aiohttp.web import Request
+
+from .claude_hooks import ClaudeHooksAPI
+from .claude_hooks import PendingToolSpan
+from .claude_hooks import SessionState
+from .claude_hooks import _format_span_id
+from .claude_hooks import _format_trace_id
+from .claude_hooks import _to_json_str
+from .codex_cost_tracker import compute_openai_cost_metrics
+from .llmobs_event_platform import with_cors
+
+log = logging.getLogger(__name__)
+
+
+def _timestamp_to_ns(timestamp: str) -> int:
+    if not timestamp:
+        return int(datetime.datetime.now(datetime.timezone.utc).timestamp() * 1_000_000_000)
+    try:
+        dt = datetime.datetime.fromisoformat(timestamp.replace("Z", "+00:00"))
+        return int(dt.timestamp() * 1_000_000_000)
+    except ValueError:
+        return int(datetime.datetime.now(datetime.timezone.utc).timestamp() * 1_000_000_000)
+
+
+def _content_text(content: Any) -> str:
+    if isinstance(content, str):
+        return content
+    if not isinstance(content, list):
+        return ""
+    parts: List[str] = []
+    for item in content:
+        if not isinstance(item, dict):
+            continue
+        if item.get("type") in ("input_text", "output_text"):
+            text = item.get("text", "")
+            if text:
+                parts.append(str(text))
+    return "\n\n".join(parts)
+
+
+def _copy_messages(messages: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
+    return [dict(message) for message in messages]
+
+
+def _append_unique_message(messages: List[Dict[str, Any]], message: Dict[str, Any]) -> None:
+    if message not in messages:
+        messages.append(message)
+
+
+class CodexTurn:
+    def __init__(self, turn_id: str, trace_id: str, root_span_id: str, step_span_id: str, start_ns: int) -> None:
+        self.turn_id = turn_id
+        self.trace_id = trace_id
+        self.root_span_id = root_span_id
+        self.step_span_id = step_span_id
+        self.start_ns = start_ns
+        self.root_span_ref: Optional[Dict[str, Any]] = None
+        self.step_span_ref: Optional[Dict[str, Any]] = None
+        self.step_span_refs: Dict[str, Dict[str, Any]] = {}
+        self.step_count = 0
+        self.current_step_start_ns: Optional[int] = None
+        self.current_step_last_ns: Optional[int] = None
+        self.current_step_has_llm = False
+        self.llm_start_ns: Optional[int] = None
+        self.llm_input_messages: List[Dict[str, Any]] = []
+        self.active_llm_input_messages: Optional[List[Dict[str, Any]]] = None
+        self.llm_output_messages: List[Dict[str, Any]] = []
+        self.pending_llm_input_messages: List[Dict[str, Any]] = []
+        self.last_step_span_ref: Optional[Dict[str, Any]] = None
+        self.last_llm_span_ref: Optional[Dict[str, Any]] = None
+        self.last_ns = start_ns
+        self.closed = False
+
+
+class CodexSession:
+    def __init__(self, session_id: str, start_ns: int) -> None:
+        self.session_id = session_id
+        self.start_ns = start_ns
+        self.cwd = ""
+        self.originator = ""
+        self.cli_version = ""
+        self.model_provider = "openai"
+        self.model = ""
+        self.effort = ""
+        self.user_prompts: List[str] = []
+        self.tools_used: Set[str] = set()
+        self.pending_tools: Dict[str, PendingToolSpan] = {}
+        self.active_turn: Optional[CodexTurn] = None
+        self.completed_turns: List[str] = []
+        self.seen_records: Set[str] = set()
+
+
+class CodexConfig:
+    def __init__(self) -> None:
+        self.ml_app = os.environ.get("DD_CODEX_ML_APP", "codex")
+        self.service = self.ml_app
+        self.env = "local"
+        self.agent_name = self.ml_app
+        self.hostname = socket.gethostname()
+        self.user_handle = os.environ.get("DD_USER_HANDLE", "")
+
+
+class CodexHooksAPI:
+    """Handler for Codex session JSONL records.
+
+    Codex already writes a structured session JSONL. Lapdog tails those files
+    and posts each record here, wrapped with the Codex session id.
+    """
+
+    def __init__(self, hooks_api: ClaudeHooksAPI) -> None:
+        self._hooks_api = hooks_api
+        self._config = CodexConfig()
+        self._sessions: Dict[str, CodexSession] = {}
+        self._raw_events: List[Dict[str, Any]] = []
+        self._last_session_id = ""
+
+    def _append_span(self, span: Dict[str, Any]) -> None:
+        self._hooks_api._assembled_spans.append(span)
+
+    def _get_or_create_session(self, session_id: str, start_ns: int) -> CodexSession:
+        if session_id not in self._sessions:
+            self._sessions[session_id] = CodexSession(session_id=session_id, start_ns=start_ns)
+            if session_id not in self._hooks_api._sessions:
+                self._hooks_api._sessions[session_id] = SessionState(
+                    session_id=session_id,
+                    trace_id=_format_trace_id(),
+                    root_span_id=_format_span_id(),
+                    start_ns=start_ns,
+                )
+        return self._sessions[session_id]
+
+    def _base_tags(self, session: CodexSession) -> List[str]:
+        tags = [
+            f"ml_app:{self._config.ml_app}",
+            f"session_id:{session.session_id}",
+            f"service:{self._config.service}",
+            f"env:{self._config.env}",
+            "source:codex-jsonl",
+            "language:python",
+            f"hostname:{self._config.hostname}",
+        ]
+        if self._config.user_handle:
+            tags.append(f"user_handle:{self._config.user_handle}")
+        return tags
+
+    def _agent_manifest(self, session: CodexSession) -> Dict[str, Any]:
+        model_settings: Dict[str, Any] = {}
+        if session.effort:
+            model_settings["reasoning_effort"] = session.effort
+        return {
+            "name": self._config.agent_name,
+            "instructions": "",
+            "handoff_description": "",
+            "model": session.model,
+            "model_provider": session.model_provider,
+            "model_settings": model_settings,
+            "tools": [{"name": name} for name in sorted(session.tools_used)],
+            "handoffs": [],
+            "guardrails": [],
+        }
+
+    def _update_agent_manifest(self, session: CodexSession) -> None:
+        turn = session.active_turn
+        if turn is None or not turn.root_span_ref:
+            return
+        metadata = turn.root_span_ref["meta"].setdefault("metadata", {})
+        dd_metadata = metadata.setdefault("_dd", {})
+        dd_metadata["agent_manifest"] = self._agent_manifest(session)
+
+    def _llm_input_messages(self, session: CodexSession, turn: CodexTurn) -> List[Dict[str, Any]]:
+        if turn.llm_input_messages:
+            return _copy_messages(turn.llm_input_messages)
+        prompt = "\n\n".join(session.user_prompts)
+        if prompt:
+            return [{"role": "user", "content": prompt}]
+        return []
+
+    def _llm_input_value(self, session: CodexSession, turn: CodexTurn) -> str:
+        messages = self._llm_input_messages(session, turn)
+        if not messages:
+            return ""
+        if len(messages) == 1 and messages[0].get("role") == "user":
+            return str(messages[0].get("content", ""))
+        return _to_json_str(messages)
+
+    def _set_step_input(self, session: CodexSession, turn: CodexTurn, step: Dict[str, Any]) -> None:
+        input_value = self._llm_input_value(session, turn)
+        if input_value:
+            step["meta"]["input"]["value"] = input_value
+
+    def _mark_llm_start(self, session: CodexSession, turn: CodexTurn, start_ns: int) -> None:
+        if turn.llm_start_ns is None:
+            turn.llm_start_ns = start_ns
+        if turn.active_llm_input_messages is None:
+            turn.active_llm_input_messages = self._llm_input_messages(session, turn)
+
+    def _set_step_output(self, turn: CodexTurn, message: str) -> None:
+        if not message:
+            return
+        step = turn.step_span_ref or turn.last_step_span_ref
+        if step is not None:
+            step["meta"]["output"]["value"] = message
+
+    def _set_llm_output(self, turn: CodexTurn, message: str) -> None:
+        if not message:
+            return
+        output_message = {"role": "assistant", "content": message}
+        if turn.llm_start_ns is not None:
+            _append_unique_message(turn.llm_output_messages, output_message)
+        if turn.last_llm_span_ref is not None and not turn.last_llm_span_ref["meta"]["output"].get("messages"):
+            turn.last_llm_span_ref["meta"]["output"]["messages"] = [dict(output_message)]
+
+    def _append_llm_span(self, turn: CodexTurn, span: Dict[str, Any]) -> None:
+        insert_at: Optional[int] = None
+        for index, existing in enumerate(self._hooks_api._assembled_spans):
+            if existing.get("parent_id") != span["parent_id"]:
+                continue
+            if existing.get("meta", {}).get("span", {}).get("kind") != "tool":
+                continue
+            if existing.get("start_ns", 0) >= span["start_ns"]:
+                insert_at = index
+                break
+        if insert_at is None:
+            self._append_span(span)
+        else:
+            self._hooks_api._assembled_spans.insert(insert_at, span)
+        turn.last_llm_span_ref = span
+
+    def _start_turn(self, session: CodexSession, record: Dict[str, Any]) -> List[str]:
+        completed = self._finalize_turn(session, status="ok")
+
+        payload = record.get("payload", {})
+        timestamp_ns = _timestamp_to_ns(record.get("timestamp", ""))
+        turn_id = payload.get("turn_id") or _format_span_id()
+        session.cwd = payload.get("cwd", session.cwd)
+        session.model = payload.get("model", session.model)
+        session.effort = payload.get("effort", session.effort)
+        trace_id = _format_trace_id()
+        root_span_id = _format_span_id()
+        shared_session = self._hooks_api._sessions.get(session.session_id)
+        if shared_session is not None:
+            shared_session.trace_id = trace_id
+            shared_session.root_span_id = root_span_id
+            shared_session.start_ns = timestamp_ns
+
+        turn = CodexTurn(
+            turn_id=turn_id,
+            trace_id=trace_id,
+            root_span_id=root_span_id,
+            step_span_id="",
+            start_ns=timestamp_ns,
+        )
+        session.active_turn = turn
+        session.user_prompts = []
+        session.tools_used = set()
+        session.pending_tools = {}
+
+        root_tags = self._base_tags(session) + ["trajectory.semantic_type:turn"]
+        root_span: Dict[str, Any] = {
+            "span_id": root_span_id,
+            "trace_id": trace_id,
+            "parent_id": "undefined",
+            "name": "codex-request",
+            "status": "ok",
+            "start_ns": timestamp_ns,
+            "duration": 0,
+            "ml_app": self._config.ml_app,
+            "service": self._config.service,
+            "env": self._config.env,
+            "session_id": session.session_id,
+            "tags": root_tags,
+            "meta": {
+                "span": {"kind": "agent"},
+                "input": {"value": ""},
+                "output": {"value": ""},
+                "model_name": session.model,
+                "model_provider": session.model_provider,
+                "metadata": {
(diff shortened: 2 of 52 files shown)


ddapm_test_agent/codex_cost_tracker.py
```diff
@@ -0,0 +1,76 @@
...
+
+@dataclass(frozen=True)
+class _OpenAIPrice:
+    prefix: str
+    input: int
+    cached_input: int
+    output: int
+
+
+_PRICING: List[_OpenAIPrice] = [
```
tomshen, line 25: probably worth checking these (also we should decide if want to hard-code these)

tomshen:
Ported three correctness fixes from the canonical trajectory Codex integration (`DataDog/trajectory/trajectory/codex/`). Redaction was the fourth gap but is handled at event intake, so it's out of scope here.

### 1. Cursor persistence (`lapdog/codex_cursor.py` + watcher wiring)
Mirrors `trajectory/codex/watcher/cursor.go`. Atomic temp+rename, lenient load (missing/corrupt → empty state, never raises). Watcher loads on startup, saves periodically (rate-limited ≥1s) and on shutdown, **prefers persisted offsets** over the 5-minute mtime heuristic, and **resets offset to 0 on file truncation**. Default location `~/.lapdog/codex-cursor.json` via new `--cursor-path` flag. The heuristic still governs the first-ever startup before a cursor exists.

### 2. Tool-use-id deduplication (`codex_hooks.py`)
Codex reuses `call_id`s like `web_search_2` across turns; today the second `function_call` silently overwrites `pending_tools[call_id]` and the first output lands nowhere. `_deduplicate_tool_use_id()` ports trajectory's `mapper.go:deduplicateToolUseID` — first occurrence keeps the id, repeats get `_2`, `_3`, … suffixed. `CodexSession` gains `seen_tool_use_ids` (counter) and `tool_use_id_map` (orig → in-flight unique). `_handle_function_call_output` resolves through the map. Cleared at turn boundary.

### 3. Subagent + compaction handling (`codex_hooks.py`)
Both event types fell through the dispatch's catch-all; now wired through. Pattern mirrors `claude_hooks.py`:
- **Subagents** (`collab_agent_spawn_begin`/`end`): emit `meta.span.kind="agent"` spans with subagent metadata (nickname/role/child_session_id/status). `agent_span_stack` + `pending_subagents` on `CodexSession`. Tool spans inside the spawn range nest under the subagent via new `_tool_parent_id` helper. Out-of-order ends unwind intervening entries; unterminated subagents close at `_finalize_turn`.
- **Compaction** (top-level `type=compacted` **and** `event_msg` subtype `context_compacted`): annotates `meta.metadata._dd.compactions` on the current active span, identical to `claude_hooks._handle_pre_compact`. No new span kind.

### Tests
20 new tests, all 44 codex-related tests pass:
- `test_codex_watcher.py`: cursor missing/corrupt/roundtrip, no stray .tmp, `_drain_file` truncation reset, end-to-end watcher resume from a pre-written cursor.
- `test_codex_hooks.py`: duplicate `call_id` produces two distinct tool spans with correct inputs/outputs, subagent emits `kind=agent` span with tool nested underneath, both compaction trigger paths annotate `_dd.compactions`.

Claude session: `f38c78cc-cf63-4fbd-9c8d-4d552ff692ef`
Resume: `claude --resume f38c78cc-cf63-4fbd-9c8d-4d552ff692ef`

tomshen:
Summary: committed and pushed Codex/Lapdog tracing reliability updates. This adds task_started turn creation and duplicate session suppression, groups child thread spans under parent sessions, keeps proxy LLM/tool spans correlated by proxy session key and step, gates X-DDAPM-Upstream overrides behind DD_CODEX_ALLOW_UPSTREAM_OVERRIDE, and improves the Codex watcher cwd/session-dir/offset handling.

Tests: python -m pytest tests/test_codex_hooks.py tests/test_codex_proxy.py tests/test_codex_watcher.py tests/test_lapdog_cli.py (69 passed).

lapdog/codex_watcher.py
```diff
@@ -0,0 +1,390 @@
...
+            state.offset = line_end
+            continue
+
+        session_id = _record_session_id(record)
+        if session_id:
+            state.session_id = session_id
+
+        record_cwd = _record_cwd(record)
+        if record_cwd and state.matches_cwd is None:
+            state.matches_cwd = os.path.abspath(record_cwd) == os.path.abspath(cwd)
```
mr-lee: **Bug — cwd filter is exact-match; regression vs Go reference.**

```python
state.matches_cwd = os.path.abspath(record_cwd) == os.path.abspath(cwd)
```

The Go reference (`trajectory/codex/watcher/watcher.go:431-442`, `isUnderProjectRoot`) does a canonicalized prefix match. If `lapdog codex` is launched from `~/repo` and Codex starts (or `chdir`s) into `~/repo/src`, this comparison fails and the entire session is silently dropped (matches_cwd=False clears the buffer at line 172). Same issue at line 214 in `_prime_file_state`.

Two sub-bugs in one line:
1. **Exact match** should be **prefix match** (`cwd` is a project root, not a single allowed directory).
2. **`abspath` doesn't resolve symlinks.** On macOS, `/tmp` → `/private/tmp` will cause spurious mismatches. Use `os.path.realpath` (or `Path.resolve(strict=False)`) and a path-segment-aware prefix check (don't just `startswith`, that matches `/foo/bar` against `/foo/barbaz`).

Suggested:
```python
def _is_under(cwd: str, root: str) -> bool:
    if not cwd or not root:
        return False
    cwd_r = os.path.realpath(cwd)
    root_r = os.path.realpath(root)
    return cwd_r == root_r or cwd_r.startswith(root_r + os.sep)
```

lapdog/codex_watcher.py
```diff
@@ -0,0 +1,390 @@
...
+        complete_data = data
+    else:
+        complete_end = data.rfind(b"\n") + 1
+        if complete_end <= 0:
+            return None
+        complete_data = data[:complete_end]
+
+    posted_session_id: Optional[str] = None
+    cursor = state.offset
+    for raw_line in complete_data.splitlines(keepends=True):
```
mr-lee: **Bug — `splitlines(keepends=True)` splits on `\r\n`, `\v`, `\f`, NEL, etc. inside JSON string values.**

JSONL is `\n`-only, but `bytes.splitlines` treats `\r`, `\r\n`, `\v`, `\f`, `\x85` (NEL), and `\x1c`-`\x1e` as line breaks. Repro:

```python
>>> b'{"a":"line1\r\nline2"}\n{"b":1}\n'.splitlines(keepends=True)
[b'{"a":"line1\r\n', b'line2"}\n', b'{"b":1}\n']
```

Any tool result containing Windows-style file content, a captured `git log`, or a model response with an embedded `\r` will be fragmented. The first fragment fails `json.loads` and is silently skipped (line 154-156); the second fragment is junk and also skipped. `state.offset` advances past both, so the record is permanently lost.

The Go reference uses `bufio.ReadBytes('\n')` + `bytes.TrimRight(lineBytes, "\n")` — splits only on `\n`. Fix: `data.split(b'\n')` and re-add the trailing `\n` to each chunk for offset accounting, or read line-by-line in binary mode with `BufferedReader.readline()`.

lapdog/codex_watcher.py
```diff
@@ -0,0 +1,390 @@
...
+        session_id = _record_session_id(record)
+        if session_id:
+            state.session_id = session_id
+
+        record_cwd = _record_cwd(record)
+        if record_cwd and state.matches_cwd is None:
+            state.matches_cwd = os.path.abspath(record_cwd) == os.path.abspath(cwd)
+
+        if state.matches_cwd is None:
+            state.buffer.append(record)
```
mr-lee: **Bug — crash window: cursor advances past in-memory-buffered records before delivery.**

When `matches_cwd is None` (first lines before a cwd-bearing record arrives), records are buffered in `state.buffer` (line 167) AND `state.offset = line_end` is advanced (line 168). `_maybe_save_cursor` persists that offset every poll cycle (line 337).

If the watcher crashes (or `lapdog codex` is SIGKILL'd, or its parent OOMs) after the cursor save but before the buffer is flushed downstream, those records are lost AND the cursor will skip past them on next start.

The Go reference doesn't have this gap — `dispatchEntry` calls `handler.OnEvent` inline; there's no "buffer pending cwd resolution" state because cwd filtering happens *after* the mapper sees the event.

Fix: don't advance `state.offset` past records still sitting in `state.buffer`. Track `state.offset_pending` (advanced as we scan) separately from `state.offset` (advanced only after successful delivery or definitive discard). Persist the latter.

lapdog/codex_watcher.py
```diff
@@ -0,0 +1,390 @@
...
+
+    return posted_session_id
+
+
+def _prime_file_state(path: Path, state: FileState, cwd: str, up_to: int) -> None:
+    """Read old records only to learn session id and cwd; do not post them."""
+    try:
+        with path.open("r") as f:
+            while True:
+                if f.tell() >= up_to:
```
mr-lee: **Bug (latent on non-ASCII) — text-mode `tell()` returns an opaque cookie, not a byte offset.**

```python
with path.open("r") as f:
    while True:
        if f.tell() >= up_to:   # up_to is from stat.st_size — a byte offset
            break
```

Python docs are explicit: text-mode `tell()` is an opaque cookie usable only with `seek`. CPython happens to return byte offsets for ASCII and most UTF-8 content today (I verified — 14 bytes of UTF-8 with multi-byte chars returned tell=14), but this is implementation-dependent and breaks under: BOMs, encoding errors triggering decoder recovery, `newline=''` mode, alternative interpreters (PyPy).

Open in `"rb"` and decode each line individually, or use `f.buffer.tell()` if you keep text mode. This also makes `_prime_file_state` consistent with `_drain_file`'s binary read, eliminating the asymmetry where one path raises `UnicodeDecodeError` on bad bytes and the other quietly succeeds.

mr-lee (commented):
Review with the perspective of two parallel codebases that have already lived through this kind of watcher:.

The three correctness fixes you ported (cursor persistence, tool-use-id dedup, subagent + compaction handling) look right and are well-tested. The findings below are divergences from the Go reference and resilience/perf concerns from those parallel codebases. Three of these are observable bugs: cwd exact-match (#1), CRLF-in-JSON-string splitting (#2), and a crash window past buffered records (#3).

I've pulled this down locally and am implementing fixes + tests for the bugs and the cheaper resilience items. Will share a diff or follow-up PR once green.

### Reference for context
- Trajectory's perf-driven move from polling to fsnotify
- Trajectory's 64KB→10MB JSONL buffer lesson + pre-commit Check 25
- EMFILE-crashes-Electron-main fix
- OOM-from-unbounded-backfill streaming fix

lapdog/codex_watcher.py
```diff
@@ -0,0 +1,390 @@
...
+            flush=True,
+        )
+        state.offset = 0
+        state.session_id = ""
+        state.matches_cwd = None
+        state.buffer.clear()
+    try:
+        with path.open("rb") as f:
+            f.seek(state.offset)
+            data = f.read()
```
mr-lee: **Perf — `f.read()` reads from offset to EOF in a single bytes allocation.**

If the gap between polls is 200MB (one big tool result + a long compaction summary across a multi-second pause), that's a 200MB bytes object allocated and immediately processed.

The Go reference uses `bufio.NewReaderSize(f, 64*1024)` and `ReadBytes('\n')` — bounded. `claude-adapter` and `gemini-adapter` explicitly stream JSONL in 64KB chunks (commit `c18d71f22f`, "OOM from unbounded backfill queries").

Trajectory also enforces a per-line cap via pre-commit Check 25: `bufio.Scanner.Buffer(64KB, 10MB)`. Apply the same: stream line-by-line via `io.BufferedReader.readline()` with a reasonable max-line cap (say 10MB).

lapdog/codex_watcher.py
```diff
@@ -0,0 +1,390 @@
...
+        if session_id:
+            state.session_id = session_id
+
+        record_cwd = _record_cwd(record)
+        if record_cwd and state.matches_cwd is None:
+            state.matches_cwd = os.path.abspath(record_cwd) == os.path.abspath(cwd)
+
+        if state.matches_cwd is None:
+            state.buffer.append(record)
+            state.offset = line_end
```
mr-lee: **Perf/correctness — `state.buffer` has no cap; unbounded growth if cwd is never determined.**

In practice `session_meta` is line 1 of every rollout, so this rarely matters. But if Codex ever changes its rollout format, or emits a long preamble of cwd-less events first, this is a silent OOM.

Cap the buffer (e.g., 1000 entries or 1MB serialised). On overflow either flush as `matches_cwd=False` with a warning, or hard-stop and emit an error event.

ddapm_test_agent/codex_hooks.py
```diff
@@ -0,0 +1,1464 @@
...
+        self._append_span(span)
+        entry: Dict[str, Any] = {
+            "span_id": span_id,
+            "parent_id": parent_id,
+            "call_id": call_id,
+            "start_ns": start_ns,
+            "agent_name": str(agent_name),
+            "_span_ref": span,
+        }
+        session.pending_subagents[call_id] = entry
```
mr-lee, line 1446: **Possible bug — `pending_subagents` keyed by raw `call_id` while tool-use-ids are deduplicated.**

The PR (correctly) handles tool-call-id reuse via `_deduplicate_tool_use_id`. But this line stores the subagent entry under the raw `call_id`, not a deduped key. If Codex reuses a `collab_agent_spawn_begin` `call_id` across turns (the same way it reuses tool call_ids), the second `begin` overwrites the first entry, and the first `end` would pop the wrong entry — causing the wrong subagent to "complete" and unwinding the stack incorrectly.

Two options:
1. Verify with the Codex team that subagent `call_id`s are globally unique across turns. If so, add a regression test that asserts two subagent spawns with id `agent_1`, `agent_1` in successive turns are treated correctly.
2. Route subagent `call_id`s through the same `_deduplicate_tool_use_id` flow.

lapdog/codex_watcher.py
```diff
@@ -0,0 +1,390 @@
...
+        except OSError:
+            pass
+    print(
+        "lapdog codex watcher: ready " f"cwd={cwd} parent_pid={parent_pid} session_dir={session_dir}",
+        file=sys.stderr,
+        flush=True,
+    )
+
+    while True:
+        for path in _iter_jsonl_files(session_dir):
```
mr-lee: **Perf — 250ms polling + recursive `rglob` is the exact pattern trajectory's `SPEC-SERVE-PERFORMANCE.md` was written to eliminate.**

At 4Hz, `_iter_jsonl_files` walks the entire session tree and stats every file. Heavy users have thousands of historical rollouts. Trajectory measured 9 stale serve processes consuming 68% CPU from ticker loops doing similar walks (commit `45334f08`).

Scope-wise this watcher is one process per `lapdog codex` invocation rather than long-lived daemons, so absolute cost is lower. But a user with many concurrent lapdog sessions or a deep `~/.codex` tree will feel it.

Two cheap wins available without a full fsnotify migration:
1. **Split discovery from drain.** rglob for new files every 1–2s; drain known `states` every 250ms. Lower-latency reads for known sessions, dramatically less FS pressure.
2. **Watch only today's date subdir** when possible (Codex sessions are date-bucketed in newer versions).

Full fix (matches trajectory): `watchdog` (Python's fsnotify wrapper) primary, polling fallback. Worth filing as a follow-up.

lapdog/codex_watcher.py
```diff
@@ -0,0 +1,390 @@
...
+    session_id: str,
+    record: Dict[str, Any],
+    source_path: Path,
+    proxy_session_key: Optional[str] = None,
+) -> bool:
+    body = {"session_id": session_id, "record": record, "source_path": str(source_path)}
+    if proxy_session_key:
+        body["proxy_session_key"] = proxy_session_key
+    try:
+        response = requests.post(
```
mr-lee: **Perf/resilience — synchronous `requests.post` blocks the watch loop; no connection reuse.**

Each event posts blocking with a 2s timeout. No `requests.Session` for TCP/TLS reuse. If lapdog's hook endpoint stalls, the whole poll cycle stalls; backed-up rollout files don't get drained until the stall resolves. With multiple active sessions, this serialises across files.

Cheap fixes:
1. Construct a `requests.Session()` once at startup and reuse it (TCP/TLS reuse, 10x latency improvement on localhost POST loops).
2. Split timeout into `(connect, read)` — e.g., `(0.5, 2.0)`. 500ms connect is generous for localhost.
3. On first POST failure for a file, break out of the inner record loop so other files in this poll cycle don't pay the timeout cost.

Bigger fix: an in-process queue with a worker thread, so file reads aren't held up by network latency. Trajectory's `2026-02-23-72-http-spy-deadlock-ddtrace.md` retro is a cautionary tale on sync I/O in hot paths.

lapdog/codex_watcher.py
```diff
@@ -0,0 +1,390 @@
...
+        return Path(explicit_session_dir).expanduser()
+    codex_home = os.environ.get("CODEX_HOME")
+    if codex_home:
+        return Path(codex_home).expanduser() / "sessions"
+    return Path.home() / ".codex" / "sessions"
+
+
+def _process_exists(pid: int) -> bool:
+    try:
+        os.kill(pid, 0)
```
mr-lee: **Resilience — PID-reuse race.**

`os.kill(parent_pid, 0)` only tells you *some* process has that PID, not necessarily the original `lapdog codex`. On a busy box PID reuse within `flush_seconds` (2s default) is rare but possible, and the watcher would never exit.

Record `psutil.Process(pid).create_time()` at launch in the parent and pass it through to the watcher; verify the PID still has that same start time before treating it as alive.

lapdog/codex_watcher.py
```diff
@@ -0,0 +1,390 @@
...
+                    states[path].ignored = True
+                if initial_offset and not states[path].ignored:
+                    _prime_file_state(path, states[path], cwd, initial_offset)
+            if states[path].ignored:
+                try:
+                    stat = path.stat()
+                except OSError:
+                    continue
+                if stat.st_size < states[path].offset:
+                    states[path] = FileState(offset=0)
```
mr-lee: **Bug — truncation of an `ignored` file silently re-enables capture from offset 0.**

```python
if stat.st_size < states[path].offset:
    states[path] = FileState(offset=0)   # new FileState defaults ignored=False
```

The new `FileState` defaults `ignored=False`. The file flips from "advance offset but don't post" to "fully capture from the start," potentially mis-attributing the replayed records to whichever `proxy_session_key` the watcher was launched with.

Fix: preserve `ignored=True` on truncation, OR drop the entry entirely (next iteration will re-evaluate).

lapdog/codex_watcher.py
```diff
@@ -0,0 +1,390 @@
...
+    if stat_mtime >= started_at - replay_recent_seconds:
+        return 0
+    return stat_size
+
+
+CURSOR_SAVE_INTERVAL_SECONDS = 1.0
+
+
+def _sync_cursor(cursor: CursorState, states: Dict[Path, FileState]) -> None:
+    cursor.files = {str(path): state.offset for path, state in states.items()}
```
mr-lee: **Subtle — cursor `files` map is replaced from `states`, not merged.**

```python
def _sync_cursor(cursor: CursorState, states: Dict[Path, FileState]) -> None:
    cursor.files = {str(path): state.offset for path, state in states.items()}
```

This wipes the cursor entry for any file not currently tracked in `states`. In steady state this is fine (files are in `states` once seen), but if `_iter_jsonl_files` transiently misses a file (race with creation, permission flip, intermittent I/O), its cursor entry vanishes for one cycle and the next discovery treats it as new — falling through to the mtime heuristic.

The Go reference (`watcher.go:496-509`) copies `w.cursor.Files` directly rather than rebuilding it from active sessions; entries are only deleted explicitly. Consider the same: merge into the loaded map, only prune on explicit deletion.

lapdog/codex_watcher.py
```diff
@@ -0,0 +1,390 @@
...
+                flush=True,
+            )
+            return False
+    except Exception as exc:
+        print(
+            f"lapdog codex watcher: hook post failed session_id={session_id} source_path={source_path}: {exc}",
+            file=sys.stderr,
+            flush=True,
+        )
+        return False
```
mr-lee, line 175: **Resilience — narrow exception catch.**

`requests.post` is wrapped in `except Exception` (good), but the watcher loop catches only `OSError` and `json.JSONDecodeError`. Not covered:
- `UnicodeDecodeError` from text-mode `_prime_file_state` on non-UTF8 bytes
- `MemoryError` from the unbounded `f.read()` (#5)
- Generic exceptions from `path.stat()` on some FS edge cases

We learned the same lesson the hard way: commit `14674fd040` "prevent EMFILE FSWatcher errors from crashing main process". The blast radius here is smaller (a subprocess vs Electron main), but a silent crash mid-session loses all subsequent events with no surface to the user.

Wrap the inner loop in a broader `except Exception` that logs to stderr and continues, at least for the per-file processing.

lapdog/codex_watcher.py
```diff
@@ -0,0 +1,390 @@
...
+
+
+def watch_codex_sessions(
+    lapdog_url: str,
+    cwd: str,
+    parent_pid: int,
+    session_dir: Path,
+    poll_interval: float = 0.25,
+    flush_seconds: float = 2.0,
+    replay_recent_seconds: float = RECENT_SESSION_REPLAY_SECONDS,
```
mr-lee, line 550: **Nit — 5-minute replay window for cursor-less files silently drops longer-running sessions.**

If a user runs a long Codex session, exits lapdog, and restarts it 6+ minutes later, the still-active session is silently skipped (mtime is older than `started_at - 300s`). Two options:

1. Widen the default to ~24h (matches trajectory's "active sessions during day" assumption).
2. Default to "always replay if cursor entry missing"; treat the heuristic as opt-out via flag.

At minimum, document the 5-min limit clearly in `--help`.

mr-lee:
Main review notes below.

**Findings**

1. **High: real Codex `turn_context` records can split a single turn into a bogus empty trace.** The PR treats every `turn_context` as a turn start in `ddapm_test_agent/codex_hooks.py:1406`, then invents a turn id when the payload has none in `ddapm_test_agent/codex_hooks.py:535`. Real trajectory fixtures model `turn_context` as state-only and put the real `turn_id` on `task_started`. In that ordering, the user prompt lands on the synthetic turn, then `task_started` finalizes it and starts the real turn with no input.

2. **High: cursor resume is bypassed in the actual `lapdog codex` path.** `cmd_codex` always starts the watcher with a proxy key in `lapdog/cli.py:504`, and proxy-key mode marks all files present at watcher startup as ignored in `lapdog/codex_watcher.py:273` and `lapdog/codex_watcher.py:315`. That happens even when a persisted cursor offset exists, then the ignored branch advances to EOF in `lapdog/codex_watcher.py:319`. Watcher restarts or resumed Codex sessions can silently skip unprocessed records.

3. **High: failed or unsupported proxy responses can suppress the JSONL fallback LLM span.** The proxy reserves an LLM call before upstream success in `ddapm_test_agent/codex_proxy.py:321`, but only registers spans for successful parseable 200 responses in `ddapm_test_agent/codex_proxy.py:401`. Later, token counts are dropped whenever `proxy_llm_calls_started > proxy_llm_usage_events_seen` in `ddapm_test_agent/codex_hooks.py:930`. A 500, parse failure, empty output, or unrecognized stream shape can lose the LLM span entirely.

4. **High: streaming disconnects continue consuming and buffering upstream output.** On downstream disconnect, `_handle_streaming` sets `downstream_closed` but keeps reading and appending chunks in `ddapm_test_agent/codex_proxy.py:361`. It then still reconstructs and registers a completed span in `ddapm_test_agent/codex_proxy.py:380`. A canceled Codex request can keep burning tokens, grow memory, and report output the client never received.

5. **High: grouped child-session proxy calls can attach to the parent turn.** `begin_proxy_llm_call()` resolves by proxy key but returns the grouped `session.session_id` in `ddapm_test_agent/codex_hooks.py:967`. After `_set_session_group` maps child sessions to the parent in `ddapm_test_agent/codex_hooks.py:1216`, `register_proxy_llm_span()` can re-resolve that parent id through `ddapm_test_agent/codex_hooks.py:945`. Subagent LLM spans can land under the parent trace instead of the child turn.

6. **Medium: several real Codex event and tool types are silently dropped.** `_handle_response_item` only handles `function_call`, `function_call_output`, `reasoning`, and assistant `message` in `ddapm_test_agent/codex_hooks.py:819`. Trajectory handles `web_search_call`, `custom_tool_call`, `custom_tool_call_output`, patch apply events, and approval events. This will miss apply-patch/custom-tool spans and permission signals in real Codex sessions.

7. **Medium: buffered pre-metadata records can be checkpointed before successful delivery.** While waiting for cwd/session metadata, `_drain_file` buffers records but advances `state.offset` in `lapdog/codex_watcher.py:166`. If later posting that buffer fails, it returns in `lapdog/codex_watcher.py:179`, and the main loop may persist the advanced cursor in `lapdog/codex_watcher.py:337`. A restart can permanently skip the unposted buffered records.

8. **Medium: watcher performance and concurrency do not carry over the trajectory lessons.** The watcher recursively scans every `*.jsonl` file in `lapdog/codex_watcher.py:219` on a 250ms loop in `lapdog/codex_watcher.py:289`. Multiple `lapdog codex` processes also share one cursor file, and `save_cursor_atomic` uses a fixed temp path in `lapdog/codex_cursor.py:54`. This can become expensive on large session trees and race between concurrent watchers.

9. **Medium: watcher shutdown does not finalize active traces.** Parent death is detected in `lapdog/codex_watcher.py:339`, but shutdown only drains files and saves cursor in `lapdog/codex_watcher.py:349`. Forwarding only happens for completed traces in `ddapm_test_agent/codex_hooks.py:1450`. If Codex exits without `task_complete` or `turn_aborted`, the last active trace can remain local and unforwarded.

**Tests**

Ran `pytest tests/test_codex_hooks.py tests/test_codex_proxy.py tests/test_codex_watcher.py tests/test_lapdog_cli.py -q`: 70 passed, with one Requests dependency warning.

mr-lee:
Pushed fixes directly to this branch rather than as a separate PR — fast-forward only, no force-push. New head: `d2c898b`. Suggest a `git pull --rebase` before your next push.

## What landed (14 commits)

Each commit addresses an inline review thread above. Each is its own atomic change so they can be reverted or reordered individually if you disagree.

| Commit | Addresses |
|---|---|
| [`ceefd1a`](../commit/ceefd1a) `fix(codex-watcher): use realpath prefix match for cwd filter` | Review #1 (cwd exact-match regression vs Go ref) |
| [`b47db84`](../commit/b47db84) `fix(codex-watcher): split JSONL on \n only` | Review #2 (`splitlines` splits on `\r\n` inside JSON strings) |
| [`07b37bc`](../commit/07b37bc) `fix(codex-watcher): do not advance durable cursor past buffered records` | Review #3 (crash window past buffered records) |
| [`188d1e9`](../commit/188d1e9) `fix(codex-watcher): read prime pass in binary mode with byte offsets` | Review #4 (text-mode `tell()` vs byte offset) |
| [`e3e91d7`](../commit/e3e91d7) `fix(codex-watcher): stream drain line-by-line with 10MB per-line cap` | Review #5 (unbounded `f.read()`) |
| [`b939c9d`](../commit/b939c9d) `fix(codex-watcher): cap FileState.buffer at 1000 records` | Review #6 (unbounded buffer growth) |
| [`b7aefb8`](../commit/b7aefb8) `test(codex-hooks): regression coverage for reused subagent call_ids` | Review #7 (verified — see notes below) |
| [`263c5d7`](../commit/263c5d7) `fix(codex-watcher): split file discovery from per-file drain` | Review #8 (FS pressure from 4Hz rglob) |
| [`c8f889b`](../commit/c8f889b) `fix(codex-watcher): pool HTTP connections via requests.Session` | Review #9 (sync `requests.post`, no connection reuse) |
| [`e91f278`](../commit/e91f278) `fix(codex-watcher): verify parent PID start time when psutil available` | Review #10 (PID-reuse race) |
| [`697d03a`](../commit/697d03a) `test(codex-watcher): pin 'ignored' flag through truncation reset` | Review #11 (see notes) |
| [`84c4b08`](../commit/84c4b08) `fix(codex-watcher): merge cursor state rather than replace it` | Review #12 (cursor entries wiped on transient stat miss) |
| [`d25f551`](../commit/d25f551) `fix(codex-watcher): catch per-file errors so one corrupt rollout cannot stop the watcher` | Review #13 (broaden exception handling) |
| [`d2c898b`](../commit/d2c898b) `chore(codex-watcher): drop unused type-ignore comment and fix indent` | Lint cleanup |

## Notable implementation choices worth your eyes

- **Fix 3** introduces `FileState.scan_offset` (advances as we read) separately from `FileState.offset` (the durable cursor, advances only on delivery or definitive discard). A post failure rewinds `scan_offset` to `offset` so retries re-read from the durable cursor — no record can be persisted-past while still buffered.
- **Fix 5** uses `readline(MAX_LINE_BYTES + 1)`: returns `MAX_LINE_BYTES + 1` bytes when a line exceeds the cap. We detect this, seek past the next `\n`, log to stderr, and continue. If the oversized line extends past EOF the pass bails and retries next poll.
- **Fix 8** new flag: `--discovery-interval` (default 1.0s). `--poll-interval` (default 0.25s) keeps existing semantics for the per-file drain loop. Backward compatible.
- **Fix 9** flips the test patch target from `requests.post` to `_session.post` (18 call sites updated). Also splits timeout into `(0.5, 2.0)` connect/read.
- **Fix 10** new flag: `--parent-start-time` (float, default `None`). When set, watcher exits if the PID exists but with a different `create_time` — defeats PID reuse. Requires `psutil`; if not installed, falls back to the existing `os.kill(pid, 0)` behavior. `psutil` is not in `setup.cfg` and adding it felt out of scope for a review-fix PR — flag if you want it added.

## Fixes 7 and 11 are test-only

- **Fix 7 (`pending_subagents` dedup):** I verified by writing two regression tests that the existing code already handles `call_id` reuse correctly. Across turns, `_finalize_turn` clears `pending_subagents`; within a turn, `spawn_end` pops the entry before the next `spawn_begin` reuses the key. No code change needed; the tests lock the guarantee so a future refactor can't regress it.
- **Fix 11 (preserve `ignored` on truncation):** Already implemented as part of Fix 8's discovery/drain split. Committed only the regression test to lock the behavior in.

## Test fixture update

`test_drain_file_drops_non_matching_cwd` had its fixture updated because the old assertion encoded the buggy exact-match cwd behavior that Fix 1 corrects. It now uses a true sibling directory (`/tmp/x` vs `/tmp/y`) so it still asserts non-matching semantics. Noted in the Fix 1 commit.

## Test status

- Before: 70 passing.
- After: 94 passing (added 24 new tests).
- No test skipped, weakened, or `xfail`'d.
- `flake8` clean on all touched files. `mypy` clean on `lapdog/codex_watcher.py` and `lapdog/codex_cursor.py`.

Happy to revert anything you disagree with — push back hard. Sorry for the volume; was easier to ship the fixes than re-explain each one.

ddapm_test_agent/codex_cost_tracker.py
Kyle-Verhoog: i think it's ok to hardcode - alternative is we move/override the cost in the static page

mr-lee:
Pushed eba4cda to address Kyle's Codex span feedback. Changes cover: bounded Codex LLM/tool message payloads to avoid freezing the LLM span view, tool status propagation/error marking, reasoning metadata linked onto preceding tool calls/tool spans, and verbatim/code display hints for tool output.\n\nValidation:\n- pytest tests/test_codex_hooks.py tests/test_codex_proxy.py -q -> 51 passed\n- git diff --check -> clean\n- pytest tests -q -> blocked locally by standalone test-agent startup failures because OTLP gRPC cannot bind [::]:4317; targeted Codex suites are green.

mr-lee:
Pushed f67f9cd to address the live badge/status tracking issue.

What changed:
- Codex root spans now keep duration=0 while a turn is active, matching the web-ui LLM Observability contract for the Running/Live badge.
- Finalized turns still replace the root duration with the completed elapsed duration.
- Pending tool calls are marked failed if a turn ends/aborts before tool output arrives, and completed/failed tool statuses are propagated into tool-call messages.

Validation:
- pytest tests/test_codex_hooks.py tests/test_codex_proxy.py tests/test_llmobs_event_platform.py -q => 92 passed
- git diff --check => clean

I also added a UI-facing regression assertion through the llmobs query-rewriter list endpoint and trace detail endpoint, so this covers the static-app path that reads custom.duration/root duration for the badge.

Kyle-Verhoog (approved):

PR #355, opened 2026-05-15
feat(lapdog): add Gemini CLI hook support

mr-lee:
## Summary
- add Gemini CLI command-hook ingestion and LLMObs span assembly
- register Gemini hook routes, including trajectory-compatible /capture/gemini/{EventType}
- add lapdog gemini launcher support that generates and installs a Gemini extension for the selected local port
- document Gemini hook usage

## Validation
- pytest tests/test_gemini_hooks.py tests/test_lapdog_gemini_extension.py tests/test_pi_hooks.py tests/test_lapdog_claude_hooks.py -q
- pytest tests/test_gemini_hooks.py tests/test_pi_hooks.py tests/test_claude_hooks.py tests/test_claude_steps.py tests/test_lapdog_claude_hooks.py -q
- python -m flake8 ddapm_test_agent/gemini_hooks.py lapdog/cli.py tests/test_gemini_hooks.py tests/test_lapdog_gemini_extension.py
- git diff --check

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,13 @@ The test agent can receive [Claude Code hook](https://docs.claude.com/en/docs/cl
 See [`.claude/setup-hooks.md`](.claude/setup-hooks.md) for full setup instructions (also usable as a Claude Code prompt to automate the setup).
 
 
+### Gemini CLI Hooks
+
+The test agent can receive Gemini CLI command-hook events and assemble them into LLM Observability traces.
+
+Run `lapdog gemini` to start the local agent, install/update the bundled Gemini extension, and launch `gemini`. The extension posts Gemini hook payloads to `/gemini/hooks/{EventType}` and uses fail-open `curl ... || true` commands so Gemini is not blocked if the local agent is unavailable.
+
+
 ### Snapshot testing
 
 The test agent provides a form of [characterization testing](https://en.wikipedia.org/wiki/Characterization_test) which
--- a/ddapm_test_agent/_clock.py
+++ b/ddapm_test_agent/_clock.py
@@ -29,3 +29,8 @@ def monotonic_wall_ns() -> int:
             t = _last_ns + 1
         _last_ns = t
         return t
+
+
+def finalized_duration_ns(start_ns: int, end_ns: int) -> int:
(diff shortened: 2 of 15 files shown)


mr-lee:
Pushed e5f5a6d to make the Gemini PR explicitly cover the live badge/status behavior.

The Gemini implementation already had the right lifecycle: BeforeAgent creates the root span with duration=0, and AfterAgent finalizes it with elapsed duration. I added regression coverage against the same static-app-facing paths used by LLM Observability:
- internal Gemini spans keep the active root duration at 0
- query-rewriter event list returns custom.duration=0 for the active root span
- trace detail returns root span duration=0 while active
- after AfterAgent, event list returns the finalized positive duration

Validation:
- pytest tests/test_gemini_hooks.py tests/test_lapdog_gemini_extension.py tests/test_llmobs_event_platform.py -q => 49 passed
- git diff --check => clean

mr-lee:
Updated this PR with an additional live-status fix in commit 814a34b.\n\nWhat changed:\n- Kept active Gemini root spans at duration 0 for the LLM Obs live badge.\n- Ensured finalized Gemini root/step/tool spans get a positive duration even if finalization happens in the same clock tick, so they exit the live/running UI state.\n- Applied the same finalized-duration helper to Claude paths too, since the previous CI run exposed the same issue in tests/test_claude_steps.py::test_stop_finalizes_open_step on Windows.\n- Added a Gemini same-tick regression through the static-app-facing list endpoint.\n\nValidation:\n- pytest tests/test_claude_steps.py::test_stop_finalizes_open_step -q\n- pytest tests/test_gemini_hooks.py::test_gemini_same_tick_finalization_exits_live_badge -q\n- pytest tests/test_claude_hooks.py tests/test_claude_steps.py tests/test_gemini_hooks.py tests/test_lapdog_gemini_extension.py tests/test_llmobs_event_platform.py -q => 79 passed\n- git diff --check\n\nI also tried the broad local test command, but this machine still fails to start the test agent because OTLP gRPC cannot bind [::]:4317, which cascades into setup errors; the focused suites above cover the changed code paths.

PR #353, opened 2026-05-14
feat(lapdog): ship Claude Code plugin via in-repo marketplace

mr-lee:
## Summary

- Add a one-plugin Claude Code marketplace at the repo root (`.claude-plugin/marketplace.json`, name `lapdog`) and a `lapdog` plugin under `plugins/lapdog/` whose `hooks/hooks.json` registers the same non-blocking `curl` hook for every Claude Code event that `lapdog/hooks.py` writes today.
- Flip `lapdog claude` hook-writing from default-on (`--disable-claude-code-hooks` opt-out) to default-off (`--hooks` opt-in). The plugin is now the recommended integration path; `lapdog claude` no longer mutates `~/.claude/settings.json` unless explicitly asked.
- Update `lapdog/README.md` with a "Claude Code plugin" install section, the new flag semantics, and a `claude plugin uninstall` step in the uninstall flow.
- Add release note covering both the feature and the flag rename (upgrade impact).

Install once the PR merges:

```bash
claude plugin marketplace add DataDog/dd-apm-test-agent
claude plugin install lapdog@lapdog
```

## Test plan

- [x] `claude plugin validate <repo>` — marketplace passes.
- [x] `claude plugin validate plugins/lapdog` — plugin passes (one expected `author` warning, intentionally suppressed).
- [x] Plugin `hooks.json` event list and command equal `_CLAUDE_CODE_EVENTS` / `_CLAUDE_CODE_HOOK` from `lapdog/hooks.py` (verified by script).
- [x] `lapdog claude --hooks` writes settings; bare `lapdog claude` does not (verified via `_maybe_write_claude_hooks` smoke test).
- [x] Removed `--disable-claude-code-hooks` flag is rejected by the parser.
- [x] Existing `tests/test_lapdog_claude_hooks.py` still aligns — it tests `write_claude_code_hooks` directly, which is unchanged.
- [ ] Reviewer: run `claude plugin marketplace add DataDog/dd-apm-test-agent` against this branch and confirm `claude plugin install lapdog@lapdog` works end-to-end against a running `lapdog start`.

PR #356, opened 2026-05-18
feat(lapdog): auto-install Claude Code plugin on `lapdog claude`

Kyle-Verhoog:
After #353, plain `lapdog claude` launches Claude **uninstrumented** unless the user has separately run `claude plugin marketplace add ...` and `claude plugin install lapdog@lapdog`. This PR closes that gap: `lapdog claude` auto-installs the plugin on first run, detecting prior installs via `~/.claude/plugins/installed_plugins.json` so the steady-state path is a no-op. Install failures print the manual commands and continue launching Claude rather than blocking the session.

With the plugin path now default, the legacy `--hooks` flag and `lapdog/hooks.py` are removed — one integration path, not two.

### Flags

- `--no-plugin-install` — skip the auto-install (Claude launches uninstrumented).

Claude session: `0930aa05-6424-4ab1-99a2-e418cfc4df50`
Resume: `claude --resume 0930aa05-6424-4ab1-99a2-e418cfc4df50`


lapdog/README.md
```diff
@@ -138,23 +138,34 @@ Code plugin, vended from a marketplace inside this same repository. The plugin
...
 modify `~/.claude/settings.json`. `claude plugin uninstall lapdog@lapdog`
 fully removes it.
 
-If you cannot or do not want to install the plugin, `lapdog --hooks claude`
-writes the equivalent hook entries directly into `~/.claude/settings.json`.
-This is opt-in: by default `lapdog claude` no longer touches your Claude Code
-settings.
+To skip the auto-install (e.g. on a locked-down machine), run
+`lapdog --no-plugin-install claude` — Claude will launch but events will not
+be captured. If the auto-install fails (no network, etc.) lapdog prints the
```
sabrenner: this is not technically true, we will still capture LLM calls via the bootstrap. this could be acceptable (ie, maybe the user just wants token counts w/o any other noise?). to create a seamless experience, i think we should consider either

1. hard blocking our llm bootstrapping by checking the existence of the plugin in the interceptor we have. although, then what value does `lapdog` even have 😛 
2. fix up the `lapdog` app to not display the sessions column / properly render it out for non-agent spans (? not entirely sure if that would be the fix) - as the llm spans generated in this case do not show up under sessions, only traces/spans, and can be disjoint, and show the empty/loading sessions navigation bar.

Image

sabrenner (approved):
code lgtm, works on my machine! also tested with `--no-plugin-install`, which is how i came to the observation below, but idt this should block landing the pr. i think most folks will opt to use the plugin. just something to consider for the best all-around experience in a follow-up.

lapdog/README.md
Kyle-Verhoog: oo yeah i think we should do 2. I noticed it as well separate from coding agent instrumentation. Any app that doesn't have session ids gets that blank side panel.

PR #357, opened 2026-05-18
feat(lapdog): add --backfill for claude, codex, and pi sessions

tomshen:
## Summary

Adds `lapdog --backfill claude`, `lapdog --backfill codex`, and `lapdog --backfill pi` for one-shot ingestion of historical agent sessions from disk. Backfilled spans carry real source timestamps, split into **one trace per user prompt**, include **token usage + estimated cost**, and land in the same `ml_app` bucket as live spans.

- Codex routes through the existing `/codex/hooks` endpoint — the watcher already handles historical rollout records.
- Claude and Pi use new `POST /claude/hooks/backfill_session` and `POST /pi/hooks/backfill_session` endpoints that build full LLMObs spans server-side. The live hooks layer only carries lifecycle events, so trying to force the LLM-call payload through it produces empty spans with wrong timestamps and no cost — hence the new endpoints.
- If the local agent stops responding mid-backfill, the client aborts cleanly with one message instead of spamming connection-refused per session.
- `--forward` is forced off during backfill so historical sessions never stream to Datadog.

Claude session: `22df68e3-635d-4121-a35b-4c1dfd61a0c8`
Resume: `claude --resume 22df68e3-635d-4121-a35b-4c1dfd61a0c8`


ddapm_test_agent/claude_backfill.py
```diff
@@ -0,0 +1,428 @@
...
+  * One LLM span per ``type=assistant`` entry, with model + token usage +
+    cost computed via ``claude_cost_tracker.compute_cost_metrics``.
+  * One tool span per tool_use / tool_result pair, parented to the LLM span
+    that requested it.
+
+Timestamps come from the entries' ``timestamp`` field (ISO-8601). Durations
+between adjacent entries are used as best-effort span durations.
+"""
+
+import datetime as _dt
```
sabrenner: nit: i think we can just do this (and update the use of it in `_parse_iso_ns`)
```suggestion
from datetime import datetime
```

ddapm_test_agent/claude_backfill.py
```diff
@@ -0,0 +1,428 @@
...
+from typing import Any
+from typing import Dict
+from typing import List
+from typing import Optional
+
+from ddapm_test_agent.claude_cost_tracker import compute_cost_metrics
+
+# Import from the live module so the ml_app value stays in sync — backfilled
+# spans must land in the same ml_app bucket as live ones.
+from ddapm_test_agent.claude_hooks import _ML_APP  # noqa: E402
```
sabrenner: maybe we could just move this constant to its own file so we don't have to do this weird importing from the claude hooks module thing. but also i think this should be fine without the `noqa` part, so maybe just try removing it.

ddapm_test_agent/claude_backfill.py
```diff
@@ -0,0 +1,428 @@
...
+        "tags": _common_tags(session_id, source="claude-code-backfill-llm"),
+        "meta": {
+            "span": {"kind": "llm"},
+            "kind": "llm",
+            "model_name": model,
+            "model_provider": "anthropic",
+            # Input messages aren't recoverable from the on-disk transcript
+            # (it only stores the assistant response, not the request payload
+            # the SDK sent upstream). Leaving empty is honest.
+            "input": {"messages": []},
```
sabrenner, line 291: is this also a limitation with the general trajectory spec? maybe the frontend can parse the source tag or _dd metadata as set below to display some message about this limitation rather than leaving users potentially confused

ddapm_test_agent/claude_backfill.py
```diff
@@ -0,0 +1,428 @@
...
+    tool_input = pending.get("input", {}) or {}
+    start_ns = pending["start_ns"]
+    duration = max(0, end_ns - start_ns)
+    is_error = bool(tool_result_block.get("is_error", False))
+    output_value = tool_result_block.get("content", "")
+    if isinstance(output_value, list):
+        output_value = " ".join((b.get("text", "") if isinstance(b, dict) else str(b)) for b in output_value)
+    elif not isinstance(output_value, str):
+        try:
+            import json as _json
```
sabrenner: this can be a top level import

ddapm_test_agent/claude_backfill.py
```diff
@@ -0,0 +1,428 @@
...
+        "metrics": {},
+        "span_links": [],
+    }
+
+
+def _to_text(value: Any) -> str:
+    if isinstance(value, str):
+        return value
+    try:
+        import json as _json
```
sabrenner: this can be a top level import

ddapm_test_agent/claude_backfill.py
```diff
@@ -0,0 +1,428 @@
...
+    """
+    spans: List[Dict[str, Any]] = []
+    current: Optional[Dict[str, Any]] = None
+    last_assistant_ts_ns: Optional[int] = None
+    # Track the previous entry's timestamp so the LLM span can have a non-zero
+    # duration (best-effort: end-to-start of consecutive entries).
+    previous_ts_ns: Optional[int] = None
+
+    # Resolve initial model — first assistant message wins.
+    initial_model = ""
```
sabrenner: is this logic needed? the first model used could be the one used to generate the session title (lower fidelity model usually), unless that's not captured in the transcript

ddapm_test_agent/claude_backfill.py
```diff
@@ -0,0 +1,428 @@
...
+    for e in entries:
+        if e.get("type") == "assistant":
+            m = (e.get("message") or {}).get("model")
+            if isinstance(m, str) and m:
+                initial_model = m
+                break
+
+    for entry in entries:
```
sabrenner, line 437: if `initial_model` is needed, can we not combine it in this loop?

ddapm_test_agent/claude_backfill.py
```diff
@@ -0,0 +1,428 @@
...
+        ts_ns = _parse_iso_ns(entry.get("timestamp"))
+        if ts_ns is None:
+            continue
+
+        if etype == "user":
+            if isinstance(content, str) and content:
+                # New user prompt → finalize prior turn, open a new one.
+                if current is not None:
+                    _finalize_turn(current)
+                current = _new_turn(session_id, cwd, initial_model, ts_ns, content)
```
sabrenner: can we update the turn model once we hit the first assistant message in the turn, not all entires for the session?

ddapm_test_agent/claude_backfill.py
sabrenner, line 1: do we handle the subagent spans here? like `Explore` agents? it doesn't seem like we do, and wondering if that is intentional

ddapm_test_agent/claude_hooks.py
```diff
@@ -1817,10 +1817,48 @@ async def handle_raw_events(self, request: Request) -> web.Response:
...
+        except Exception:
+            return web.json_response({"error": "invalid JSON"}, status=400)
+
+        session_id = body.get("session_id") or ""
+        cwd = body.get("cwd") or ""
+        entries = body.get("entries") or []
+        if not session_id or not isinstance(entries, list):
+            return web.json_response({"error": "session_id and entries required"}, status=400)
+
+        from ddapm_test_agent import claude_backfill
```
sabrenner: this can probably be a top level import (assuming we move the `_ML_APP` constant to it's own file, otherwise we do have a circular import path)

ddapm_test_agent/pi_backfill.py
```diff
@@ -0,0 +1,378 @@
...
+import secrets
+import socket
+from typing import Any
+from typing import Dict
+from typing import List
+from typing import Optional
+
+# Import from the live module so backfilled spans land in the same ml_app
+# bucket as live ones, regardless of env-var overrides.
+from ddapm_test_agent.pi_hooks import _ML_APP  # noqa: E402
```
sabrenner: likewise, maybe this can be moved into its own file so we

ddapm_test_agent/pi_backfill.py
```diff
@@ -0,0 +1,378 @@
...
+        "service": _ML_APP,
+        "env": "local",
+        "session_id": session_id,
+        "tags": _common_tags(session_id, source="pi-backfill-llm"),
+        "meta": {
+            "span": {"kind": "llm"},
+            "kind": "llm",
+            "model_name": model,
+            "model_provider": str(msg.get("provider") or "anthropic"),
+            "input": {"messages": []},
```
sabrenner, line 262: do pi transcripts also not include input messages? maybe we can make the same kind of callout

ddapm_test_agent/pi_backfill.py
```diff
@@ -0,0 +1,378 @@
...
+        "metrics": {},
+        "span_links": [],
+    }
+
+
+def _to_text(value: Any) -> str:
+    if isinstance(value, str):
+        return value
+    try:
+        import json as _json
```
sabrenner: can be a top level import

ddapm_test_agent/pi_backfill.py
sabrenner, line 1: seeing what appears to be a lot of the same patterns between this and the claude backfill one, wondering if it's possible to DRY up the code a bit (it might not be but maybe worth a try)

lapdog/cli.py
sabrenner, line 1: for the coding agents here, would it make sense / be ok to still start the coding agents when `--backfill` is enabled?

lapdog/cli.py
```diff
@@ -508,12 +559,29 @@ def cmd_codex(sub_cmd_args: List[str], forward_data: bool) -> None:
...
+            # lapdog flag is also consumed as a lapdog arg, so the flag can
+            # appear in either position.
+            remaining = list(cmd_args[arg_idx:])
+            sub_cmd_and_args: List[str] = [remaining[0]]
+            for sub_arg in remaining[1:]:
+                if sub_arg in _LAPDOG_FLAGS:
+                    lapdog_args.append(sub_arg)
+                else:
+                    sub_cmd_and_args.append(sub_arg)
+            return lapdog_args, sub_cmd_and_args
```
sabrenner: i don't think we should do this, it muddies the behavior a bit. we should just keep
```
lapdog [...lapdog_args] [sub_cmd] [...sub_cmd_args]
```

what if, for whatever reason, the sub command has the same command flag? imho it's ergonomical enough the way we have it now. but lmk if you think there's a reason to keep it!

sabrenner (commented):
majority nits. a couple of questions/observations

1.  if someone runs `--backfill` twice, if we re-post/process the spans? i could totally be missing it but i think we do?
2. on forwarding - i see in the PR description we're not supporting it, but it might be nice to assume we want to do forwarding off the bat, if it's not too much work. although, it could be better implemented if we persist spans we create locally (so, if that's the plan, then perfect). the current random trace id/span id would not fit in well for forwarding. also worth noting that if we do decide for these commands to start up the coding agents, we will have to explicitly set `--forward` to `False`.
3. i also noticed a couple of visual things about the data & structure when QAing this, but i can follow up with those questions offline since I'm not sure what's intended in the spec.

but overall seems pretty clean and it did backfill!

ddapm_test_agent/claude_backfill.py
tomshen: Done in 1862dff: changed this to `from datetime import datetime` and updated `_parse_iso_ns` accordingly.

ddapm_test_agent/claude_backfill.py
tomshen: Done in 1862dff: moved the shared app names into `ddapm_test_agent/lapdog_app_names.py`, so backfill no longer imports the hook module just for `_ML_APP`.

ddapm_test_agent/claude_backfill.py
tomshen: Added a structured `_dd.backfill.limitations.input_messages_unavailable = true` marker on backfilled Claude LLM spans. The source tag remains too, but this gives the UI a direct signal for this limitation.

ddapm_test_agent/claude_backfill.py
tomshen: Done in 1862dff: removed this local import path by moving JSON/text serialization into shared `backfill_utils.to_text`.

ddapm_test_agent/claude_backfill.py
tomshen: Done in 1862dff: same cleanup here, now using the shared `backfill_utils.to_text` helper.

ddapm_test_agent/claude_backfill.py
tomshen: Good catch. I removed the session-wide first-assistant model pre-scan so title-generation records before the first user turn do not become the model for every turn.

ddapm_test_agent/claude_backfill.py
tomshen: Done. There is no separate `initial_model` pass now; the main loop sets the current turn's root model when it sees that turn's first assistant message.

ddapm_test_agent/claude_backfill.py
tomshen: Done in 1862dff. The turn root span and agent manifest model now update on the first assistant message within that turn.

ddapm_test_agent/claude_backfill.py
tomshen: This was missing, not intentional. I added `Task` tool_use/tool_result handling so backfilled Claude subagents show up as child `agent` spans, with a regression test.

ddapm_test_agent/claude_hooks.py
tomshen: Done: moving the app-name constants into `lapdog_app_names.py` removes the cycle, so `claude_backfill` is now imported at module scope.

ddapm_test_agent/pi_backfill.py
tomshen: Done: both Claude and Pi now read their app-name constants from `ddapm_test_agent/lapdog_app_names.py`.

ddapm_test_agent/pi_backfill.py
tomshen: Added the same `_dd.backfill.limitations.input_messages_unavailable = true` marker to Pi backfilled LLM spans.

ddapm_test_agent/pi_backfill.py
tomshen: Done via shared `backfill_utils.to_text`, so this no longer needs a function-local import.

ddapm_test_agent/pi_backfill.py
tomshen: DRY'd the small shared pieces: ID generation, JSON/text serialization, backfill metadata, duplicate-session detection, and app-name constants. I left the transcript parsing separate because the Claude and Pi formats diverge enough that combining those loops made the behavior less clear.

lapdog/cli.py
tomshen: I kept backfill from starting the coding-agent binaries; it only starts/reuses local lapdog and exits after replay. Forwarding also stays forced off: Event Platform rejects traces older than a day, and supporting only newer traces would require a separate decision about partial-session forwarding.

lapdog/cli.py
tomshen: Done in 1862dff. Parser behavior is back to `lapdog [...lapdog_args] [sub_cmd] [...sub_cmd_args]`, and I added a regression test so subcommand flags are not re-parsed as lapdog flags.

tomshen:
Pushed 1862dff with the review responses.

Summary:
- Backfill endpoints now skip a session if that `session_id` was already backfilled into the running local agent.
- Shared app-name constants and small backfill utilities remove circular/lazy import weirdness.
- Claude backfill now sets the root model from the first assistant message in each turn and renders `Task` tool calls as child agent spans for subagents.
- Claude and Pi LLM spans include structured `_dd.backfill.limitations.input_messages_unavailable` metadata for the missing input-message limitation.
- CLI parsing is back to `lapdog [...lapdog_args] [sub_cmd] [...sub_cmd_args]`, and detached startup waits until lapdog is reachable before immediate backfill POSTs.

Forwarding remains intentionally off for backfill. Event Platform rejects traces older than a day; supporting only newer traces would also require deciding whether partial sessions are acceptable.

QA:
- `python3 -m pytest tests/test_backfill_claude.py tests/test_backfill_pi.py tests/test_lapdog_cli.py`
- `env -u DD_CLAUDE_CODE_ML_APP python3 -m pytest tests/test_claude_hooks.py tests/test_pi_hooks.py tests/test_codex_hooks.py tests/test_codex_watcher.py`
- Actual isolated CLI run: `lapdog --backfill claude`, `lapdog --backfill pi`, and `lapdog --backfill codex` each forwarded one temp session into a temp local lapdog agent (`session_counts {"claude-qa": 2, "codex-qa": 1, "pi-qa": 2}`).
- `python3 -m flake8 ...` and `python3 -m mypy ...` on touched files passed.
- `python3 -m black --check ...` passed.

tomshen:
Pushed `9adddb2` with the backfill fixes.

Summary of the three issues we tried to fix:

1. Backfilled Claude/Pi sessions were not creating step spans, so the trace shape jumped directly from the root agent span to LLM/tool spans. The backfill now synthesizes `step` spans and parents LLM/tool spans under the step.
2. Claude `Task` subagents were not represented in backfill at all. Task tool results now become child `agent` spans under the relevant step, preserving the subagent prompt/result metadata.
3. Some completed backfilled sessions could get incorrect zero/short durations, which made closed sessions appear as running in Lapdog. Completed backfill spans now use a closed-duration helper with a 1ns floor, and assistant inference timing starts from the previous transcript event instead of the completion timestamp.

QA: I ran actual Claude/Pi/Codex backfills against a branch-built local test agent

ddapm_test_agent/claude_backfill.py
sabrenner, line 1: i could be missing it in an explanation somewhere, but it looks like subagents (ex. `Explore`) are still not backfilling properly.

for example, i prompted claude to spawn an explore agent, and the resulting trace structure was

Image

however, once backfilled, it looks like it got split into two different _sessions_:

Image

Image

so it still feels a bit off. not gonna block on this, but it does explode the number of sessions i think in this case.

ddapm_test_agent/claude_backfill.py
sabrenner: this is good w me!

ddapm_test_agent/pi_backfill.py
```diff
@@ -0,0 +1,458 @@
...
+_HOSTNAME = socket.gethostname()
+_USERNAME = os.environ.get("HOST_USER") or getpass.getuser()
+
+
+def _format_trace_id() -> str:
+    return format_trace_id()
+
+
+def _format_span_id() -> str:
+    return format_span_id()
```
sabrenner: we don't need functions wrapping other functions verbatim, we can just use `format_trace_id` and `format_span_id` where they are needed (and similarly in other files)

ddapm_test_agent/pi_backfill.py
```diff
@@ -0,0 +1,458 @@
...
+    """Completed backfill spans must not use duration=0.
+
+    The static LLM Observability app treats a root span with zero duration as
+    still running. Historical backfill payloads are complete by definition, so
+    use a 1ns floor when adjacent transcript entries share a timestamp.
+    """
+    return max(1, end_ns - start_ns)
+
+
+def _common_tags(session_id: str, source: str = "pi-backfill") -> List[str]:
```
sabrenner: can we use the `base_tags` from the hooks api?

lapdog/cli.py
```diff
@@ -428,8 +432,25 @@ def cmd_claude(
...
+    Claude Code transcripts from ``~/.claude/projects`` through
+    ``/claude/hooks``, and exit without launching Claude. ``forward_data``
+    is forced off and plugin installation is skipped during backfill.
+    """
+    if backfill:
+        port = _ensure_lapdog_running(forward_data=False, detached=True)
+        if port is None:
+            print("[lapdog] Could not determine lapdog port.", file=sys.stderr)
+            sys.exit(1)
+        from lapdog import backfill_claude
```
sabrenner: i think this plus the other ones (pi, codex) can just be top-level imports

sabrenner (approved):
backfilling seems to work except for the one case i identified but idt we need to block on it. left a couple of other nits as well, but non-blocking

ddapm_test_agent/pi_backfill.py
tomshen: Done in d99f50e: removed the verbatim `_format_trace_id` / `_format_span_id` wrappers in both Claude and Pi backfill and now call the shared helpers directly.

ddapm_test_agent/pi_backfill.py
tomshen: Done in d99f50e: Pi backfill now builds tags through `ClaudeHooksAPI.base_tags(...)` with the Pi ml_app/source, then appends the backfill marker. That also brings over the same project/git metadata path as live Pi hooks.

lapdog/cli.py
tomshen: Done in d99f50e: moved the Claude, Pi, and Codex backfill module imports to the top of `lapdog/cli.py`.

ddapm_test_agent/claude_backfill.py
tomshen: Yeah, this is still a real limitation. This PR stitches the parent transcript’s `Task` tool_use/tool_result into an agent span, but the directory backfill still treats separate Claude subagent transcript files as independent sessions when it walks the transcript directory. Fixing the session explosion properly would need cross-file stitching between the parent Task span and the child transcript/session, so I’m leaving that as follow-up rather than making this PR infer that relationship heuristically.

Will do this in a followup PR!

PR #362, opened 2026-05-20
fix(lapdog): return live duration for in-flight spans

clementgbcn:
## What does this PR do?

Fix `GET /claude/hooks/spans` to return a live elapsed duration for root spans of sessions that are still running (prompt submitted, `Stop` not yet fired).

## Who will it impact?

* [slack/#k9-ask-cloud-siem](https://dd.enterprise.slack.com/archives/C02PKN3LV96)

## Motivation

While a prompt is running, the lapdog dashboard showed Time & Duration as `0` because the root span is created with `duration: 0` as a placeholder and only updated when the `Stop` hook fires. The `/claude/hooks/spans` endpoint now returns `now - start_ns` for in-flight root spans without mutating the underlying span (preserving the `-1` permission-wait sentinel).

## Testing Guidelines

- Added `test_hook_in_flight_span_has_live_duration`: submits a prompt without `Stop`, queries `/claude/hooks/spans`, asserts `duration > 0`.
- All 14 existing `test_claude_hooks` tests continue to pass.

### DIFF
--- a/ddapm_test_agent/claude_hooks.py
+++ b/ddapm_test_agent/claude_hooks.py
@@ -1811,7 +1811,18 @@ class ClaudeHooksAPI:
 
     async def handle_spans(self, request: Request) -> web.Response:
         """Handle GET /claude/hooks/spans — return all assembled spans."""
-        return web.json_response({"spans": self._assembled_spans})
+        now_ns = monotonic_wall_ns()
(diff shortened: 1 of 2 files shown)


clementgbcn:
Closing it since it has been fixed in the frontend.

PR #367, opened 2026-05-22
Fix lapdog codex app watcher

tomshen:
## Summary

Fix `lapdog codex app` so Codex Desktop sessions are tailed into Lapdog like Codex CLI sessions.

Root cause: `codex app` hands off to the desktop app and exits, so the existing watcher was parented to a short-lived launcher process. It also resolved cwd only from CLI `-C/--cd`, not the app workspace path.

This PR:
- resolves `codex app [PATH]` and `codex --cd <path> app` workspace paths correctly
- keeps app-mode watchers alive by parenting them to the Lapdog agent process
- avoids duplicate app watchers for the same port/workspace with a singleton pid file
- keeps CLI-mode Codex behavior using per-launch proxy session keys

## Validation

- `python -m black --check lapdog/cli.py tests/test_lapdog_cli.py`
- `git diff --check`
- `python -m pytest tests/test_lapdog_cli.py tests/test_codex_watcher.py tests/test_codex_hooks.py tests/test_codex_proxy.py`
- Manual QA: launched `lapdog codex app` against an isolated Lapdog port, sent live Codex Desktop messages, verified they appeared in `/codex/hooks/raw`, `/claude/hooks/spans`, and the remote Lapdog UI at https://lapdog.datadoghq.com/.


lapdog/cli.py
```diff
@@ -463,19 +495,103 @@ def _resolve_codex_cwd(args: List[str]) -> str:
...
-            cwd = str(Path(cd_value).expanduser())
-            if not os.path.isabs(cwd):
-                cwd = os.path.join(os.getcwd(), cwd)
+            cwd = _abs_path_from_cwd(cd_value)
         idx = next_idx
     return os.path.abspath(cwd)
 
 
-def _start_codex_watcher(port: int, proxy_session_key: Optional[str] = None, cwd: Optional[str] = None) -> None:
+def _codex_command_index(args: List[str]) -> Optional[int]:
```
sabrenner: non-blocking - can we move some of these codex helpers into their own file? i know we have other codex files already as well. just to not bloat this file

lapdog/cli.py
```diff
@@ -445,7 +445,39 @@ def cmd_pi(sub_cmd_args: List[str], forward_data: bool) -> None:
...
+    "-a",
+    "--ask-for-approval",
+    "-c",
+    "--config",
+    "-C",
+    "--cd",
+    "--download-url",
+    "--enable",
+    "-i",
+    "--image",
```
sabrenner: tl;dr we can probably remove these two i think from this list of commands to check.

small note, i think it's fine but want to know your judgement, if someone does
```
codex -i /some/image/path app /another/image/path "explain these images"
```
as we check currently use this list of commands to parse the command index and app cwd, we could incorrectly parse `app` here as a command and not an image path. so it's probably safe to remove these two entries, although it's probably also fine to keep as `app` as an image path is probably unlikely.

lapdog/cli.py
```diff
@@ -445,7 +445,39 @@ def cmd_pi(sub_cmd_args: List[str], forward_data: bool) -> None:
...
 
 
-def _resolve_codex_cwd(args: List[str]) -> str:
+_CODEX_VALUE_FLAGS = {
+    "-a",
+    "--ask-for-approval",
+    "-c",
+    "--config",
+    "-C",
+    "--cd",
```
sabrenner: nit: looks like `--add-dir` is also an option

sabrenner (approved):
works for me now, thanks for updating! some nits but not blocking.

lapdog/cli.py
tomshen: Done in 87a1d96: moved the Codex argv parsing and app watcher key helpers into `lapdog/codex_args.py`, leaving `lapdog/cli.py` to call the small helper API.

lapdog/cli.py
tomshen: Agreed. Done in 87a1d96: removed `-i`/`--image` from the value-skip set and added a regression test covering image arguments followed by positional text named `app`, so that no longer gets treated as `codex app`.

lapdog/cli.py
tomshen: Done in 87a1d96: added `--add-dir` to the value-skip set and covered it with a resolver test.

lapdog/cli.py
```diff
@@ -445,37 +447,102 @@ def cmd_pi(sub_cmd_args: List[str], forward_data: bool) -> None:
...
-        idx = next_idx
-    return os.path.abspath(cwd)
-
-
-def _start_codex_watcher(port: int, proxy_session_key: Optional[str] = None, cwd: Optional[str] = None) -> None:
+def _codex_watcher_pid_file(log_dir: str, singleton_key: str) -> str:
+    return os.path.join(log_dir, f"codex-watcher-{singleton_key}.pid")
+
+
+def _codex_watcher_pid_from_file(pid_path: str) -> Optional[int]:
```
sabrenner: nit: we can probably re-use `_read_pid_file` instead of this new method, but have it take in an optional `path` variable, so we'd call

```python
pid, _ = _read_pid_file(path=_codex_watcher_pid_file())
```

instead of `_codex_watcher_pid_from_file()`

lapdog/cli.py
```diff
@@ -445,37 +447,102 @@ def cmd_pi(sub_cmd_args: List[str], forward_data: bool) -> None:
...
+            return int((f.readline() or "").strip())
+    except (OSError, ValueError):
+        return None
+
+
+def _codex_watcher_reusable(pid: int, parent_pid: int) -> bool:
+    if not _process_exists(pid):
+        return False
+    if os.name == "nt":
+        return True
```
sabrenner: it seems a bit arbitrary to me to always return `True` for windows, maybe just a one-line comment here explaining this could be helpful!

sabrenner:
looks like there is a codex-related failure and/or flake in CI

lapdog/cli.py
```diff
@@ -445,37 +447,102 @@ def cmd_pi(sub_cmd_args: List[str], forward_data: bool) -> None:
...
+    try:
+        result = subprocess.run(
+            ["ps", "-p", str(pid), "-o", "command="],
+            check=False,
+            capture_output=True,
+            text=True,
+            timeout=1,
+        )
+    except (OSError, subprocess.SubprocessError):
+        return True
```
sabrenner: should we actually return `False` in this case? if not, maybe another one-line explanation here could be helpful in understanding this default case.

lapdog/cli.py
```diff
@@ -445,37 +447,102 @@ def cmd_pi(sub_cmd_args: List[str], forward_data: bool) -> None:
...
+
+def _codex_watcher_pid_from_file(pid_path: str) -> Optional[int]:
+    try:
+        with open(pid_path) as f:
+            return int((f.readline() or "").strip())
+    except (OSError, ValueError):
+        return None
+
+
+def _codex_watcher_reusable(pid: int, parent_pid: int) -> bool:
```
sabrenner: in QAing this, i had codex do a review, and it raised this point:

> A stale pid file with a reused PID can make Lapdog skip starting the real watcher, or terminate an unrelated process. I’d return `False` when identity cannot be verified, or persist enough metadata to validate the watcher before reuse/kill.

i think it has merit - i honestly think it's pretty safe in its current form, but let's add a couple more safeguards if applicable to make sure we're not killing the wrong process by using this function

lapdog/cli.py
```diff
@@ -445,37 +447,102 @@ def cmd_pi(sub_cmd_args: List[str], forward_data: bool) -> None:
...
+    if "lapdog.codex_watcher" not in parts:
+        return False
+    try:
+        parent_idx = parts.index("--parent-pid")
+    except ValueError:
+        return False
+    return parent_idx + 1 < len(parts) and parts[parent_idx + 1] == str(parent_pid)
+
+
+def _stop_codex_watcher_singleton(singleton_key: str, parent_pid: int) -> None:
```
sabrenner: i think this and the above method could do with a couple of docstrings, just because the process management is a lil tricky and is non-standard based on our other hook/plugin-based integrations here

i do think the methods are readable, but just for documented intent

lapdog/cli.py
```diff
@@ -553,8 +625,23 @@ def cmd_codex(sub_cmd_args: List[str], forward_data: bool) -> None:
     if port is None:
         print("[lapdog] Could not determine lapdog port.", file=sys.stderr)
         sys.exit(1)
-    proxy_session_key = uuid.uuid4().hex
-    _start_codex_watcher(port, proxy_session_key=proxy_session_key, cwd=_resolve_codex_cwd(sub_cmd_args))
+    app_mode = codex_args.is_app_command(sub_cmd_args)
+    proxy_session_key = None if app_mode else uuid.uuid4().hex
+    parent_pid = os.getpid()
+    if app_mode:
```
sabrenner, line 752: possibly valid codex review comment:

> `cmd_codex` only stops the legacy key for the currently opened app cwd on line 635, then starts/reuses the new all-cwd singleton on line 642. If a user has legacy cwd-keyed watchers for other workspaces, those keep running and the new all-cwd watcher can duplicate posts for those sessions. Since this branch explicitly migrates from cwd-keyed to all-cwd app watchers, consider scanning/removing other verified codex-watcher-*.pid legacy watchers during app startup.

sabrenner (approved):
new updates lgtm too, just a couple of comments. 

the release note & pr description highlight the goal well, but i think a few comments in some places i've pointed out could help maintainability a bit since this is slightly different from our other coding agent integrations.

also included a couple codex review comments i think have some merit, when I was QAing these changes 😛

lapdog/cli.py
tomshen: Done in 9dc1703: `_read_pid_file` now accepts an optional `path`, and the watcher pid-file reads use that instead of a separate `_codex_watcher_pid_from_file` helper.

lapdog/cli.py
tomshen: Done in 9dc1703: removed the unconditional Windows `True`. Windows now verifies the process command line via PowerShell/CIM and returns `False` when identity cannot be checked. Added a Windows-path regression test for this.

lapdog/cli.py
tomshen: Done in 9dc1703: this now returns `False` when the watcher command line cannot be read, instead of treating an unverifiable PID as reusable.

lapdog/cli.py
tomshen: Done in 9dc1703: watcher reuse/termination now validates the process command line, parent pid, Lapdog URL, and all-cwd mode before acting on a pid-file PID. Stale same-key watcher processes are terminated only after command-line metadata matches, and unverifiable live PIDs are left alone rather than killed.

lapdog/cli.py
tomshen: Done in 9dc1703: added docstrings around the watcher matching/reuse and legacy watcher cleanup helpers to document the process-management intent.

lapdog/cli.py
tomshen: Done in 9dc1703: app startup now scans `codex-watcher-*.pid` files and stops verified legacy cwd-keyed app watchers, while skipping the new all-cwd singleton. The cleanup only terminates processes that match the expected watcher metadata.

tomshen:
Fixed the Windows Codex test failure in 9dc1703. Root cause was `_codex_watcher_reusable` returning `True` unconditionally on Windows, so the parent-pid assertion failed there. It now verifies the watcher command line on Windows too and returns `False` when the PID cannot be verified.

Validation run locally:
- `python -m black --check lapdog/cli.py tests/test_lapdog_cli.py`
- `git diff --check`
- `python -m pytest tests/test_lapdog_cli.py`
- `python -m pytest tests/test_lapdog_cli.py tests/test_codex_watcher.py tests/test_codex_hooks.py tests/test_codex_proxy.py`
- actual `python -m lapdog.cli --forward codex app` launch reused the current all-workspace watcher and left a single watcher process running.

tomshen:
Follow-up on CI: after the watcher hardening change, CI exposed a test-only timing flake in `tests/test_claude_steps.py::test_step_without_tools` on Windows. Fixed in `f03cf99` by making the test helper use the same monotonic wall clock as the production proxy path.

Latest PR checks are green across the full macOS/Ubuntu/Windows Python matrix, including the previously failing Windows 3.9/3.11/3.12 lanes.

PR #370, opened 2026-05-25
feat(agent): extract LLMObs spans from v0.4 trace meta_struct

mabdinur:
## Summary

dd-trace-py PR [#18254](https://github.com/DataDog/dd-trace-py/pull/18254) ships LLMObs on kept `APM_AGENT_PROXY` traces via `meta_struct["_llmobs"]` on `/v0.4/traces` instead of posting to `/evp_proxy/.../llmobs`. system-tests still assert on EVP-shaped session requests, so those tests go empty without agent-side bridging.

This change makes the test agent treat trace-embedded LLMObs like real EVP intake:

- **`GET /test/session/requests`** appends synthetic `POST /evp_proxy/v4/api/v2/llmobs` entries (SDK envelope JSON, not UI-remapped) so `llmobs_requests()` / `wait_for_llmobs_requests()` work unchanged.
- **`get_llmobs_spans`** also reads v0.4 traces and merges UI-formatted spans into query-rewriter / lapdog results.

## End state

After merge and a new test-agent release, INTEGRATION_FRAMEWORKS LLMObs tests can run against a tracer on the agent-based path without changing system-tests assertion code. Bump `agent_image` in system-tests once a tagged image is published.

## Edge cases

- **Dedup by `(trace_id, span_id)`** when the same span appears both as a real EVP post (predicted-drop fallback writer) and in trace `meta_struct` — synthesis is skipped; EVP wins in `get_llmobs_spans`.
- **v0.4 only** — v0.5 has no `meta_struct` slot; v0.7 does not run `_flexible_decode_meta_struct` today.
- **Malformed / partial payloads** — missing `trace_id`, span ids, or non-dict `_llmobs` values are skipped without failing the session or list endpoints.

## Reviewer notes

- Synthetic session bodies intentionally skip `remap_sdk_span_to_ui_format`; system-tests expect SDK shape.
- `_dd.trace_id` / `apm_trace_id` are 32-char hex from the APM span’s integer `trace_id`, not the LLMObs `trace_id` string in meta_struct.
- Dedup ordering in `get_llmobs_spans`: EVP-derived spans are collected first so they win over trace rebuilds.

## Test plan

- [x] `pytest` on new/updated tests (`test_llmobs_trace_extract`, session-requests, event-platform list, meta_struct round-trip)
- [ ] Build/publish test-agent image and re-run failing system-tests (INTEGRATION_FRAMEWORKS anthropic LLMObs) — planned follow-up

Made with [Cursor](https://cursor.com)


mabdinur:
@Yun-Kim I am using this system test PR to validate this fix: https://github.com/DataDog/system-tests/pull/7047/changes

ddapm_test_agent/llmobs_trace.py
```diff
@@ -0,0 +1,123 @@
...
+    apm_span_id = span.get("span_id")
+    apm_trace_id = span.get("trace_id")
+    if llmobs_trace_id is None or apm_span_id is None or apm_trace_id is None:
+        return None
+
+    _dd_attrs: Dict[str, Any] = dict(llmobs_data.get("_dd") or {})
+    try:
+        apm_trace_id_hex = format(int(apm_trace_id), "032x")
+    except (TypeError, ValueError):
+        apm_trace_id_hex = str(apm_trace_id)
```
Yun-Kim: Just in case there's a trace ID > 64 bits, this might not work. Here's how we do it on the APM processing backend to remap LLMO span events from meta_struct: https://github.com/DataDog/dd-go/blob/prod/trace/apps/trace-indexer/processors/llmobs_spans/processor.go#L190-L195

Yun-Kim (approved):

ddapm_test_agent/llmobs_trace.py
mabdinur: Good catch. Fixed in 3770e77: when a trace is 128-bit, the v0.4 span only carries the low 64 bits in `trace_id`, so we now recombine them with the high 64 bits from the `_dd.p.tid` meta tag ((high << 64) | low) before rendering the 32-char hex, matching the backend remapping. Added a unit test covering the `_dd.p.tid` case.

Issue #366, opened 2026-05-22
Safari blocks https://lapdog.datadoghq.com local access to http://localhost:8126


DavidSouther:
Safari Version 26.3.1 (21623.2.7.111.2) blocks requests to lapdog's local runner.

```
! Fetch API cannot load http://localhost:8126/info due to access control checks.
! XMLHttpRequest cannot load http://localhost:8126/api/v1/logs-analytics/aggregate?type=llmobs due to access control checks.
```

DavidSouther:
In Chrome, a permission window asks the user to allow this. Perhaps I missed this in Safari?

Kyle-Verhoog:
👀 i am also not seeing it working in safari. We'll look into this @DavidSouther - thanks for reporting!

PR #372, opened 2026-05-26
fix(lapdog): fix miscellaneous issues (OSError, README)

sabrenner:
Fixes a few user-facing things:
- should resolve the intermittent `OSError: [Errno 22] Invalid argument` logs coming through that I don't believe were surfacing as actual issues, but cleans up noise from logs nonetheless (& adds a test, which could possibly be flaky as a subprocess test, so as a reference, can be removed in the future if needed)
- README update to remove `leash` references

Non user-facing things:
- Cleans up a flake with a duration check on Windows tests by using an existing source code helper to generate wall clock times properly


tomshen (approved):

DavidSouther:
Any way to bring /leash back? I would ask to reconsider in light of https://github.com/DataDog/dd-apm-test-agent/issues/366 and general "Hey I'd love to keep this whole thing local" security vibes. (But I'm also not a ddawg customer ;) )

sabrenner:
@DavidSouther `lapdog` (at https://lapdog.datadoghq.com) is completely local (local in the sense that it uses local data and does not forward it to Datadog) unless you have a Datadog account and enter an API key to forward data! But by default, that behavior is off, and data stays local and is ephemeral, going away once you run `lapdog stop`:

Screenshot 2026-05-26 at 4 43 19 PM

For transparency, `/leash` has not been implemented at all and was a mis-reference in our docs. https://lapdog.datadoghq.com hits `localhost:8126` directly to poll data from coding sessions prefixed with `lapdog` when run (ie `lapdog claude`). with regards to the linked issue, we should have a solution coming for that as well!

Let me know if this is not clear, we're definitely open to re-wording if needed, but hopefully this helps clear it up!

DavidSouther:
Bluntly, I don't trust that type of cross origin frame busting. There's a reason browser vendors make it difficult or impossible.

sabrenner:
@DavidSouther totally fair. We are also planning on adding a way to query these traces/spans locally as well (via `localhost:8126/lapdog/spans` or something to that effect), so feel free to watch out for that one coming soon-ish!

DavidSouther:
Of course I have no idea the feasibility of this, but just shipping the current lapdog interface in tree and at /leash, feels like the best of both worlds?

PR #373, opened 2026-05-26
Add pi model cost fallback

tomshen:
## Summary
- Prefer cost values reported directly by pi hook usage data when available.
- Fall back to model-name-based cost estimation for pi spans, including AI Gateway-style `openai/...` and `anthropic/...` model IDs.
- Add regression coverage and a release note.

## QA
- Add DD gateway models: https://github.com/ddoghq-sandbox/datadog-pi-packages/blob/main/packages/refresh-models/README.md (you'll need the fixed version in https://github.com/ddoghq-sandbox/datadog-pi-packages/pull/2)
- Check that cost shows up in Lapdog for these models
- Check for cost regressions for using models directly

### DIFF
--- a/ddapm_test_agent/pi_hooks.py
+++ b/ddapm_test_agent/pi_hooks.py
@@ -60,13 +60,10 @@ _AI_GATEWAY_COST_PROVIDERS = frozenset({"anthropic", "openai"})
 
 def _split_ai_gateway_model_id(model_id: str) -> Tuple[Optional[str], str]:
     """Return (provider, model) for AI Gateway-style IDs like ``openai/gpt-5.5``."""
-    parts = model_id.split("/")
-    for index, provider in enumerate(parts[:-1]):
-        normalized_provider = provider.lower()
-        if normalized_provider in _AI_GATEWAY_COST_PROVIDERS:
-            model = "/".join(parts[index + 1 :])
-            if model:
-                return normalized_provider, model
+    provider, _, model = model_id.partition("/")
+    normalized_provider = provider.lower()
+    if model and normalized_provider in _AI_GATEWAY_COST_PROVIDERS:
+        return normalized_provider, model
     return None, model_id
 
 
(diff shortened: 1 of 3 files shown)


ddapm_test_agent/pi_hooks.py
```diff
@@ -48,11 +49,81 @@
...
 
 log = logging.getLogger(__name__)
 
 _ML_APP = os.environ.get("DD_PI_CODING_AGENT_ML_APP", "pi-coding-agent")
+_AI_GATEWAY_COST_PROVIDERS = frozenset({"anthropic", "openai"})
+
+
+def _split_ai_gateway_model_id(model_id: str) -> Tuple[Optional[str], str]:
+    """Return (provider, model) for AI Gateway-style IDs like ``openai/gpt-5.5``."""
+    parts = model_id.split("/")
```
sabrenner: nit: any chance we can just strip the first thing? that way we don't have to re-join at the end:
```python
maybe_provider, model_id_rest = model_id.split("/", 1)
if maybe_provider.lower() in _AI_GATEWAY_COST_PROVIDERS:
    ...
```

or are there more cases which i'm not thinking of? makes this method a little less roundabout

ddapm_test_agent/pi_hooks.py
```diff
@@ -48,11 +49,81 @@
...
+        model_id=pricing_model_id,
+        non_cached_input_tokens=non_cached_input_tokens,
+        cache_write_tokens=cache_write_tokens,
+        cache_read_tokens=cache_read_tokens,
+        output_tokens=output_tokens,
+    )
+    if anthropic_cost is not None:
+        return anthropic_cost
+
+    if _looks_like_openai_model(pricing_model_id):
```
sabrenner: could we move this check to the above one or does it need to be here? ie
```python
if provider == "openai" or _looks_like_openai_model(pricing_model_id):
    ...
```

sabrenner (approved):
couple of questions but lgtm!

ddapm_test_agent/pi_hooks.py
tomshen: Agreed. Simplified this in 5aed25c to split only the leading provider segment and leave other IDs unchanged.

ddapm_test_agent/pi_hooks.py
tomshen: Done in 5aed25c. OpenAI providers and GPT-looking model IDs now share the same branch before Anthropic pricing.

PR #374, opened 2026-05-26
Capture coding agent project metadata

tomshen:
## Summary

- Add shared coding-agent project metadata resolution for Lapdog traces, including normalized `git.repository_url` and `project_name` fallback behavior

## QA
- Test for codex, claude, and pi


ddapm_test_agent/claude_hooks.py
```diff
@@ -350,6 +357,41 @@ def _set_hidden_metadata(self, span: Dict[str, Any], **kwargs: Any) -> None:
...
+            f"source:{source}",
+            "language:python",
+            f"hostname:{_HOSTNAME}",
+        ]
+        if _USER_HANDLE:
+            tags.append(f"user_handle:{_USER_HANDLE}")
+        tags.extend(project_metadata_tags(session.project_metadata))
+        return tags
+
+    def _apply_project_metadata_to_span(self, session: SessionState, span: Dict[str, Any]) -> None:
```
sabrenner: do we need a function wrapping another function that basically just passes parameters through?

ddapm_test_agent/claude_hooks.py
```diff
@@ -350,6 +357,41 @@ def _set_hidden_metadata(self, span: Dict[str, Any], **kwargs: Any) -> None:
         """Merge key-value pairs into span['meta']['metadata']['_dd'], preserving existing values."""
         span["meta"].setdefault("metadata", {}).setdefault("_dd", {}).update(kwargs)
 
+    def _update_session_project_metadata(self, session: SessionState, body: Dict[str, Any]) -> None:
```
sabrenner: is there a good way to verify updating the cwd of a running session?

ddapm_test_agent/claude_hooks.py
```diff
@@ -350,6 +357,41 @@ def _set_hidden_metadata(self, span: Dict[str, Any], **kwargs: Any) -> None:
...
+            f"session_id:{session.session_id}",
+            f"service:{_ML_APP}",
+            "env:local",
+            f"source:{source}",
+            "language:python",
+            f"hostname:{_HOSTNAME}",
+        ]
+        if _USER_HANDLE:
+            tags.append(f"user_handle:{_USER_HANDLE}")
+        tags.extend(project_metadata_tags(session.project_metadata))
```
sabrenner, line 397: if we gate this check behind `if session`, then callers don't have to care about session having a value (i saw claude_proxy doing a `_base_tags if session else [...]` kinda thing which we don't have to do if we do this). we can throw the `session_id` tag in that check as well, and leave the base tags as-is.

ddapm_test_agent/codex_hooks.py
```diff
@@ -411,8 +417,28 @@ def _base_tags(self, session: CodexSession, source: str = "codex-jsonl") -> List
...
+        project_name = extract_agent_project_name(payload) or extract_agent_project_name(record)
+        git_repository_url = extract_git_repository_url(record) or extract_git_repository_url(payload)
+        session.project_metadata = resolve_project_metadata(
+            cwd=session.cwd,
+            project_name=project_name or ("" if cwd_changed else session.project_metadata.project_name),
+            git_repository_url=git_repository_url
+            or ("" if cwd_changed else session.project_metadata.git_repository_url),
+        )
+
+    def _apply_project_metadata_to_span(self, session: CodexSession, span: Dict[str, Any]) -> None:
```
sabrenner: similar nit about not re-wrapping functions

ddapm_test_agent/pi_hooks.py
```diff
@@ -309,6 +311,24 @@ def _current_parent_id(self, session: SessionState) -> str:
     def _append_span(self, span: Dict[str, Any]) -> None:
         self._hooks_api._assembled_spans.append(span)
 
+    def _base_tags(self, session: SessionState, source: str = "pi-hooks") -> List[str]:
```
sabrenner: can we not re-use the `_base_tags` from the `_hooks_api`? better yet, could this be a general util function? seems it doesn't rely on `self` at all.

ddapm_test_agent/pi_hooks.py
```diff
@@ -309,6 +311,24 @@ def _current_parent_id(self, session: SessionState) -> str:
...
+            f"source:{source}",
+            "language:python",
+            f"hostname:{_HOSTNAME}",
+        ]
+        if _USER_HANDLE:
+            tags.append(f"user_handle:{_USER_HANDLE}")
+        tags.extend(project_metadata_tags(session.project_metadata))
+        return tags
+
+    def _apply_project_metadata_to_span(self, session: SessionState, span: Dict[str, Any]) -> None:
```
sabrenner: same nit about not re-wrapping functions

releasenotes/notes/coding-agent-project-metadata-55e245541f94453f.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    Capture git repository and project name metadata for coding agent sessions captured by lapdog.
```
sabrenner:
```suggestion
    Lapdog: Capture git repository and project name metadata for coding agent sessions captured by lapdog.
```

nit on prefixing release notes for clarity

tests/test_claude_hooks.py
```diff
@@ -114,6 +115,56 @@ async def test_hook_tool_use_creates_tool_span(agent):
     assert tool["parent_id"] == root_spans[0]["span_id"]
 
 
+async def test_claude_project_metadata_from_hook_cwd(agent, tmp_path, monkeypatch):
+    monkeypatch.delenv("DD_GIT_REPOSITORY_URL", raising=False)
+    subprocess.run(["git", "init"], cwd=tmp_path, check=True, capture_output=True)
+    subprocess.run(
```
sabrenner, line 124: i think it's fine to leave here but wondering if it could be worth making this a fixture instead with some kinda cleanup?

sabrenner (commented):
mostly nits but lgtm! i QA'd with the three coding agents, although i wasn't sure how to test out the cwd changing, so i'll approve after that. aren't sessions scoped to a specific dir off the bat? i saw i could add more dirs during a session, but that did not seem to change the git tags added

(i could just be oblivious to new/existing changes 😭)

ddapm_test_agent/pi_hooks.py
sabrenner: comment is outdated now but the new diff also seems to re-define the function

ddapm_test_agent/claude_hooks.py
tomshen: Dropped in 4a10c1d.

ddapm_test_agent/claude_hooks.py
tomshen: Added a test in 4a10c1d (`test_claude_project_metadata_updates_when_cwd_changes`) — sends two `UserPromptSubmit`/`Stop` turns with different `cwd` values pointing at two distinct git repos, asserts each turn's root span gets that turn's project metadata. Also QA'd live against a running lapdog by POSTing two turns at different cwds.

On your other question — `/add-dir` in Claude Code adds dirs to the *allowlist* but doesn't change the agent's `cwd`, which is why the tags didn't move. Claude/Pi/Codex all send `cwd` on every hook event, so a real cwd change (e.g. process restart in a different repo, or codex `turn_context` with a new cwd) is what triggers re-resolution.

ddapm_test_agent/claude_hooks.py
tomshen: Done in 4a10c1d — `_create_llm_span` now builds the base tag list unconditionally and only appends `session_id` / `user_handle` / project tags inside `if session:`. The ternary is gone. Same pattern in `_adopt_orphan_spans` (orphans get session-scoped tags appended on adoption rather than rebuilt).

ddapm_test_agent/codex_hooks.py
tomshen: Dropped in 4a10c1d.

ddapm_test_agent/pi_hooks.py
tomshen: Deduped in 4a10c1d — pi_hooks now delegates: `self._hooks_api.base_tags(session, source="pi-hooks", ml_app=_ML_APP)`. Added an `ml_app` kwarg to `ClaudeHooksAPI.base_tags` so pi can override its own ml_app (different env var default in non-lapdog mode). Resisted promoting it to a coding_agent_metadata util to avoid 6-kwarg call sites — happy to do it that way if you prefer.

ddapm_test_agent/pi_hooks.py
tomshen: Dropped in 4a10c1d.

releasenotes/notes/coding-agent-project-metadata-55e245541f94453f.yaml
tomshen: Applied in 4a10c1d.

tests/test_claude_hooks.py
tomshen: Leaving as-is for now — there are only three tests that need the cache cleared and inlining the call sites is one line each. Happy to extract to a fixture if more callers show up.

ddapm_test_agent/pi_hooks.py
tomshen: Fixed in 4a10c1d — pi_hooks no longer defines its own `base_tags` and delegates to `_hooks_api.base_tags`.

tomshen:
@sabrenner re: "aren't sessions scoped to a specific dir off the bat?" — yes, in Claude Code the session is scoped to the cwd where you launched `claude`, and `/add-dir` doesn't change that:

- Claude Code's hooks send `cwd` on every event, and it stays equal to the launch directory for the life of the session. `/add-dir` only widens the *filesystem allowlist* so tools can read paths outside the launch dir — it doesn't `chdir`. The hook payload's `cwd` doesn't move, so the git tags don't either. That matches what you saw.
- Codex is the agent where cwd actually *can* change mid-session: its `turn_context` JSONL records include a per-turn `cwd`, so if codex changes dir between turns we re-resolve and the next root span picks up new tags. The new `test_claude_project_metadata_updates_when_cwd_changes` test simulates this with two `UserPromptSubmit`/`Stop` turns at different cwds.
- Pi is similar: the lapdog extension grabs `cwd` from `pi.sessionManager.getCwd()` on each event.

So in practice with Claude Code you'd only see the tags change across a process restart in a different repo. The mid-session re-resolution code path is mostly there to cover codex/pi semantics and to be defensive.

ddapm_test_agent/coding_agent_metadata.py
```diff
@@ -0,0 +1,191 @@
...
+            timeout=3,
+        )
+    except subprocess.TimeoutExpired:
+        log.debug("git %s in %s timed out", " ".join(args), cwd)
+        return ""
+    except (OSError, subprocess.SubprocessError) as exc:
+        log.debug("git %s in %s failed: %s", " ".join(args), cwd, exc)
+        return ""
+    if result.returncode != 0:
+        return ""
```
sabrenner, line 190: worth having a debug log in this case too maybe?

sabrenner (approved):
lgtm!

ddapm_test_agent/coding_agent_metadata.py
tomshen: Added in 5e8dea0 — non-zero git exit now logs the return code and stderr at debug.

PR #376, opened 2026-05-27
fix(lapdog): update session model slug from API response after /model switch

### LINKED ISSUE #375: lapdog: traces show previous model after switching models with /model in Claude Code
## Description

After switching the Claude Code model using the `/model` command, lapdog continues to show the **previous** model name in all traces — including those generated after the switch.

## Steps to Reproduce

1. Start a Claude Code session with lapdog tracing enabled (default model: `claude-opus-4-7`)
2. Use `/model` to switch to a different model (e.g. `claude-sonnet-4-6`)
3. Send a few messages
4. Open the lapdog trace viewer and inspect the traces

## Expected Behavior

Traces generated after the model switch should reflect the new model (`claude-sonnet-4-6`).

## Actual Behavior

All traces — including those sent after the `/model` switch to Sonnet 4.6 — still display `claude-opus-4-7` as the model. The model metadata in trace payloads is not updated when the active model changes.

Image

## Environment

- Claude Code CLI
- lapdog trace viewer (dd-apm-test-agent)

Hasaber8:
Fixes #375. session.model was only set at session_start and never updated, so lapdog showed the original model on all traces even after switching via /model.

*** Updated ***

- claude_proxy.py: sync session.model from the Anthropic API response in _create_llm_span — the same source already used for cost calculation

image

### DIFF
--- a/ddapm_test_agent/claude_hooks.py
+++ b/ddapm_test_agent/claude_hooks.py
@@ -645,6 +645,17 @@ class ClaudeHooksAPI:
 
         log.info("Claude session started: %s (model=%s)", session_id, model)
 
+    def _handle_model_select(self, session_id: str, body: Dict[str, Any]) -> None:
+        """Handle model_select hook event — update session model when user switches via /model."""
+        session = self._get_or_create_session(session_id)
+        model_id = body.get("model_id", "") or body.get("model", "")
+        model_provider = body.get("model_provider", "")
+        if model_id:
+            session.model = model_id
+        if model_provider:
+            session.model_provider = model_provider
+        log.info("Claude model changed: %s → %s", session_id, model_id)
+
     def _finalize_interrupted_turn(self, session: SessionState) -> None:
         """Finalize an in-progress turn that was interrupted (e.g. user Ctrl+C).
 
@@ -1561,6 +1572,7 @@ class ClaudeHooksAPI:
 
         handlers: Dict[str, Any] = {
             "SessionStart": self._handle_session_start,
+            "model_select": self._handle_model_select,
             "UserPromptSubmit": self._handle_user_prompt_submit,
             "PreToolUse": self._handle_pre_tool_use,
             "PostToolUse": self._handle_post_tool_use,
--- a/ddapm_test_agent/claude_hooks.py
+++ b/ddapm_test_agent/claude_hooks.py
@@ -650,11 +650,12 @@ class ClaudeHooksAPI:
         session = self._get_or_create_session(session_id)
         model_id = body.get("model_id", "") or body.get("model", "")
         model_provider = body.get("model_provider", "")
-        if model_id:
+        previous_model = session.model
+        if model_id and model_id != "unknown":
             session.model = model_id
         if model_provider:
             session.model_provider = model_provider
-        log.info("Claude model changed: %s → %s", session_id, model_id)
+        log.info("Claude model changed: %s %s → %s", session_id, previous_model, model_id)
 
     def _finalize_interrupted_turn(self, session: SessionState) -> None:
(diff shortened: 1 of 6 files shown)


sabrenner:
@rohanhasabe-esper thanks for the bump here, and thanks for the contribution! We'll take a look at this and run CI if everything looks good

Hasaber8:
> @rohanhasabe-esper thanks for the bump here, and thanks for the contribution! We'll take a look at this and run CI if everything looks good

Works! My bad on mixing up my accounts here 😅

ddapm_test_agent/claude_hooks.py
```diff
@@ -1573,6 +1585,7 @@ def _dispatch_hook(self, body: Dict[str, Any]) -> None:
 
         handlers: Dict[str, Any] = {
             "SessionStart": self._handle_session_start,
+            "model_select": self._handle_model_select,
```
sabrenner: as far as I can tell, this hook does not exist for Claude Code, so we can remove it here.

i saw in the PR description this was supposed to be for Pi, so as a reference, we could actually put it in [this file](https://github.com/DataDog/dd-apm-test-agent/blob/master/ddapm_test_agent/pi_hooks.py). however, it seems that file already does handle `model_select`, so i think we can remove this handler from this file altogether. 

it's the below change in `claude_proxy` that does the trick for me when testing this locally for Claude Code.

tests/test_claude_hooks.py
```diff
@@ -317,6 +317,40 @@ async def test_hook_agent_manifest(agent):
     assert root["meta"]["model_provider"] == "anthropic"
 
 
+async def test_hook_model_select_updates_model(agent):
```
sabrenner: i think we can remove this test in favor of the manual test result you've uploaded as a screenshot in the PR description, and from what I've seen testing this locally. if we get rid of the hook above, this becomes hard to test because we haven't set up a good way to reliably unit test changes to `claude_proxy.py`. but in this case, the change is easy to assert manually, and we can work on getting unit tests in place for this so that we can make a regression case down the line

sabrenner (commented):
looks great! i think we can remove the `model_select` handler from the `claude_hooks.py` module, as it is not a CC hook, and we already handle it in Pi. the change to `claude_proxy` is sufficient.

once updated, I'll run CI and we can land this - thanks again for the contribution!

ddapm_test_agent/claude_hooks.py
Hasaber8: Not sure which PR you are referencing to here.

sabrenner (commented):
thanks for applying the feedback, looks good now! i forgot to mention this before, but you can generate a release note as well:

```bash
riot run reno new {release-note-slug}
```

where `release-note-slug` is just the the filename for the release note, and fill in the `fixes` section ([example](https://github.com/DataDog/dd-apm-test-agent/blob/master/releasenotes/notes/fix-session-id-association-for-claude-code-llm-spans-8793d1c132599595.yaml)), that would be great, thank you!

Hasaber8:
> thanks for applying the feedback, looks good now! i forgot to mention this before, but you can generate a release note as well:
> 
> ```shell
> riot run reno new {release-note-slug}
> ```
> 
> where `release-note-slug` is just the the filename for the release note, and fill in the `fixes` section ([example](https://github.com/DataDog/dd-apm-test-agent/blob/master/releasenotes/notes/fix-session-id-association-for-claude-code-llm-spans-8793d1c132599595.yaml)), that would be great, thank you!

Done

sabrenner (approved):

sabrenner:
Thanks for the contribution @Hasaber8! This will land in the next release, which will probably be this week 😎

Hasaber8:
@sabrenner Thank you for the tool!

PR #380, opened 2026-05-28
Lapdog: Tag coding agent spans with git.commit.sha

tomshen:
## What

Lapdog: tag coding-agent LLM Obs spans with `git.commit.sha` — the commit that is HEAD of the **session's repository** at the moment each span starts. Because the SHA flips the instant a commit lands, this captures *when commits happen* during a session and lets traces be filtered/grouped by commit.

## How

The commit SHA is resolved from the **same per-session `cwd`** that the existing `git.repository_url` tag is derived from (`resolve_project_metadata` / `_local_git_metadata`), so the two tags always describe the same repository.

- **`ddapm_test_agent/coding_agent_metadata.py`** — new `git_commit_sha_for_cwd(cwd)` (current HEAD via `git rev-parse HEAD`, TTL-cached ~1s per cwd so new commits show up promptly without spawning git per span) and `git_commit_sha_tags(cwd)`. Uses the same `cwd or os.getcwd()` resolution as `resolve_project_metadata`.
- **`claude_hooks.base_tags` / `codex_hooks._base_tags`** — append `git.commit.sha:<sha>` right where `git.repository_url` is already added. This covers Claude, Pi (shares the Claude hooks API), and Codex.

Because tagging happens at span build time (≈ span start), the SHA reflects the commit applicable at the start of each span. Spans created after a commit lands get the new SHA. Non-git working dirs / empty repos simply omit the tag (consistent with `git.repository_url`).

## Testing

- `tests/test_coding_agent_metadata.py` — resolver tracks HEAD across commits, TTL cache behavior, non-repo → empty, tag format.
- `tests/test_claude_hooks.py::test_claude_spans_tagged_with_git_commit_sha` — end-to-end: post hooks for a session whose `cwd` is a real repo, assert root + tool spans carry `git.commit.sha:<sha>` alongside `git.repository_url`.


ddapm_test_agent/coding_agent_metadata.py
```diff
@@ -104,6 +108,40 @@ def project_metadata_tags(metadata: CodingAgentProjectMetadata) -> List[str]:
...
+
+    Uses the same cwd resolution as ``resolve_project_metadata`` (falling back
+    to the process cwd when unset) so the commit SHA always describes the same
+    repository as the ``git.repository_url`` tag. Returns ``""`` when ``cwd`` is
+    not inside a git work tree.
+    """
+    cwd = str(cwd or "").strip() or os.getcwd()
+    now = (_now or time.monotonic)()
+    cached = _commit_sha_cache.get(cwd)
+    if cached is not None and now - cached[0] < _COMMIT_SHA_TTL_SECONDS:
```
sabrenner: while i think it's nice to have a cache, and the TTL is very low WRT coding agent latencies, i wonder if there could be race conditions that cause us to tag the wrong commit hash.

since that is a self-imposed race condition, i wonder if it's better to just leave it to the `_run_git` method to be the only race condition, which is sorta out of our control.

but, not blocking, just a thought!

ddapm_test_agent/coding_agent_metadata.py
```diff
@@ -104,6 +108,40 @@ def project_metadata_tags(metadata: CodingAgentProjectMetadata) -> List[str]:
...
+# unlike the immutable remote URL — it cannot live in the cached, frozen
+# CodingAgentProjectMetadata. Instead we resolve the repo's current HEAD on
+# demand, keyed on the same cwd the `git.repository_url` tag is derived from,
+# with a short TTL so new commits are reflected within ~1s without spawning git
+# for every span.
+_COMMIT_SHA_TTL_SECONDS = 1.0
+_commit_sha_cache: Dict[str, Tuple[float, str]] = {}
+
+
+def git_commit_sha_for_cwd(cwd: Any, *, _now: Optional[Callable[[], float]] = None) -> str:
```
sabrenner: it seems we don't use `_now` in source code, only in testing code. correct me if i'm wrong, but in that case, i think we can just patch the `monotonic` method for `time` with a patch fixture in tests as needed 

but, if we get rid of the cache, this won't really matter. but if we keep it, i think we should do that instead. also if we keep it, should we just use our [`_clock`](https://github.com/DataDog/dd-apm-test-agent/blob/master/ddapm_test_agent/_clock.py) module?

ddapm_test_agent/codex_hooks.py
```diff
@@ -419,6 +420,9 @@ def _base_tags(self, session: CodexSession, source: str = "codex-jsonl") -> List
         if self._config.user_handle:
             tags.append(f"user_handle:{self._config.user_handle}")
         tags.extend(project_metadata_tags(session.project_metadata))
+        # git.commit.sha: the commit that is HEAD of the same repo the
+        # git.repository_url tag describes, at the moment this span starts.
+        tags.extend(git_commit_sha_tags(session.cwd))
```
sabrenner: i forget, can this not use the `base_tags` from the other hooks file?

sabrenner (approved):
lgtm, tried with pi, codex, and claude, saw the tag and it's value change as the coding agent committed something.

one thing i noticed is that the tags are not on claude code llm spans - we probably just need to add the call to `self._hooks_api.base_tags` in `claude_proxy.py`. that should be addressed here, but otherwise everything lgtm so approving (w some other optional comments)!

ddapm_test_agent/coding_agent_metadata.py
tomshen: Fixed in `0f7fa11`: I removed the TTL cache entirely, so `git_commit_sha_for_cwd` now resolves `git rev-parse HEAD` directly and there is no self-imposed stale-cache window.

ddapm_test_agent/coding_agent_metadata.py
tomshen: Fixed in `0f7fa11`: removing the cache also removed the `_now` test hook and the cache-specific tests, so there is no clock injection here anymore.

ddapm_test_agent/codex_hooks.py
tomshen: Fixed in `0f7fa11`: `CodexHooksAPI._base_tags` now delegates to `ClaudeHooksAPI.base_tags(session, source=..., ml_app=...)`, so Codex and Claude share the project/user/git commit tag construction.

tomshen:
Addressed the review note about Claude Code LLM spans in `0f7fa11`: proxy-created Claude LLM spans now use `self._hooks_api.base_tags(session, source="claude-code-proxy")`, and adopted orphan proxy spans use the same session-scoped tags. Added `test_proxy_llm_span_uses_session_base_tags_with_git_commit_sha` to cover `git.commit.sha` on Claude proxy LLM spans.

ddapm_test_agent/codex_hooks.py
```diff
@@ -407,19 +405,7 @@ def _get_or_create_session(self, session_id: str, start_ns: int) -> CodexSession
         return self._sessions[session_id]
 
     def _base_tags(self, session: CodexSession, source: str = "codex-jsonl") -> List[str]:
```
sabrenner: nit: we don't need `_base_tags` to wrap `_hooks_api.base_tags`, we can just call it directly in all the places we currently use `_base_tags` in this file

tomshen: Fixed in `c297094`: removed `CodexHooksAPI._base_tags` and now call `self._hooks_api.base_tags(...)` directly at each Codex tag site. I added an explicit `user_handle` override to the shared helper so Codex keeps its `DD_USER_HANDLE` behavior from `CodexConfig`.

PR #381, opened 2026-05-28
Lapdog: Fix pi error spans, input-less traces, and stale turn ends

tomshen:
## What

Fixes three bugs in the pi coding-agent lapdog integration.

### Bug 1 — failed LLM calls produced an empty step (no error span)

pi surfaces a failed LLM call as an assistant message with `stopReason: "error"`
(and an `errorMessage`); in the thrown-exception path (connection refused, etc.)
it may not emit `message_end` at all. Previously the handler always built the LLM
span with `status: "ok"` and dropped the error message, so the step looked empty.

- `pi_lapdog_extension.ts` now forwards `error_message` on `message_end`.
- `pi_hooks.py` marks the LLM span (and its enclosing step) as an error with the
  provider message when `stop_reason == "error"`.
- New `_flush_orphaned_pending_llm` emits an error LLM span when a step finalizes
  with a pending LLM call that never received `message_end`, so the step is never
  silently empty.

### Bug 2 — traces created without user input

`agent.continue()` (after auto-compaction or auto-retry) emits a fresh
`agent_start` with no user input. The handler treated every `agent_start` as a new
turn, rotating to a new `trace_id` — producing traces with an empty user prompt.

- `pi_lapdog_extension.ts` marks continuation `agent_start`s via
  `is_continuation`, driven by `before_agent_start` (which only fires for genuine
  user prompts, never for `continue()`).
- `pi_hooks.py` extends the current trace on a continuation (keeps `trace_id`,
  root span, and `user_prompts`; reopens it so the next `agent_end` re-finalizes
  and re-forwards) instead of starting a new one. A `missing_user_input` metadata
  flag is added as a diagnostic for any genuine turn that still lacks a prompt.

### Bug 3 — stale `turn_end` events closed the next active turn

Pi can emit `turn_start` for the next turn before the previous turn's `turn_end`
arrives. Previously, the stale `turn_end` finalized whatever step was active,
which could close the next turn before its LLM message arrived and produce an
extra empty step span.

- `pi_hooks.py` now compares the incoming `turn_end.turn_index` with the active
  step's `turn_index` and ignores stale mismatches instead of finalizing the
  active step.

## Tests

- 6 new tests in `tests/test_pi_hooks.py` (error span, orphaned pending LLM on
  `agent_end` and `session_shutdown`, continuation merge, real-second-turn still
  rotates, stale `turn_end` handling).
- Verified end-to-end against a real `pi` 0.76.0 process driving the test agent
  (see QA below).

## QA instructions

These run the real lapdog flow with forwarding on, so the resulting spans show up
in the lapdog UI at https://lapdog.datadoghq.com.

**Prereqs**

```bash
# From this branch:
git fetch origin ts/fix-pi && git checkout ts/fix-pi
pip install -e .

# Make sure DD_API_KEY (and DD_SITE if you are not on US1) is set so lapdog can
# forward to the backend, then drop any already-running agent so the new code is used:
lapdog stop
```

`lapdog --forward pi <pi-args>` starts the forwarding agent, installs this
branch's extension, and launches pi. After each run, open
https://lapdog.datadoghq.com and find your most recent **`pi-coding-agent`**
session (match it by the time you ran the command), then open the trace.

### Repro Bug 1 — error LLM span

Force a real LLM error with a bad API key (401 is not retryable, so you get one
clean error span):

```bash
lapdog --forward pi --provider anthropic --model claude-3-5-haiku-20241022 \
  --api-key "sk-ant-invalid-key-for-testing" --no-session \
  -p "Say hello in one word"
```

In the lapdog UI, open that session's trace. **Expected:** the turn
(`pi-request`) contains a step with an **LLM span marked as an error** whose
message is the 401 (`invalid x-api-key`) detail — the step is no longer empty.
On `master` the same run shows an empty step with no error.

### Repro Bug 2 — input-less traces on continuation (auto-retry / compaction)

Stand up a local provider that always returns a retryable `503 overloaded`, so pi
retries via `agent.continue()` (each retry emits a fresh `agent_start` with no
user input). A project-local fast-retry config keeps it quick.

```bash
mkdir -p /tmp/qa && cd /tmp/qa
mkdir -p .pi
echo '{"retry":{"enabled":true,"maxRetries":2,"baseDelayMs":200}}' > .pi/settings.json

# 503-overloaded server
cat > srv.py <<'PY'
import http.server, socketserver
class H(http.server.BaseHTTPRequestHandler):
    def _e(self):
        b=b'{"type":"error","error":{"type":"overloaded_error","message":"Overloaded"}}'
        self.send_response(503); self.send_header("Content-Length",str(len(b))); self.end_headers(); self.wfile.write(b)
    def do_POST(self):
        n=int(self.headers.get("Content-Length",0));
        if n: self.rfile.read(n)
        self._e()
    def do_GET(self): self._e()
    def log_message(self,*a): pass
socketserver.TCPServer(("127.0.0.1",8199),H).serve_forever()
PY
python3 srv.py & SRV=$!

# extension registering a provider that points at the 503 server
cat > badprovider.ts <<'TS'
import type { ExtensionAPI } from "@mariozechner/pi-coding-agent";
export default function badProvider(pi: ExtensionAPI): void {
  pi.registerProvider("badgw", {
    name: "Bad Gateway (test)", baseUrl: "http://localhost:8199",
    apiKey: "test-key", api: "anthropic-messages",
    models: [{ id: "bad-model", name: "Bad Model", reasoning: false, input: ["text"],
      cost: { input: 0, output: 0, cacheRead: 0, cacheWrite: 0 }, contextWindow: 200000, maxTokens: 4096 }],
  });
}
TS

lapdog --forward pi --provider badgw --model bad-model \
  -e badprovider.ts --no-session -p "Say hello"

kill $SRV
```

In the lapdog UI, open that session. **Expected:** the first attempt and both
retry continuations are a **single trace / turn** whose input is the original
`"Say hello"` prompt (every inference shows the error). On `master` you instead
see multiple separate traces in the session, two of which have **no user input**.

### Repro Bug 3 — stale `turn_end` closes the next active turn

This is covered by `test_stale_turn_end_does_not_finalize_next_step`, which
sends `turn_start(1)` before the delayed `turn_end(0)`. **Expected:** the stale
`turn_end(0)` is ignored, turn 1 stays open until its LLM response arrives, and
no extra empty step span is emitted.

> Optional extension-payload sanity check (not shown in the UI): the running
> agent exposes the raw hook events it received at `GET /pi/hooks/raw`. With
> forwarding-on lapdog that's `curl -s http://localhost:8126/pi/hooks/raw`. You
> should see one `agent_start` with `is_continuation:false` (and the real
> `user_prompt`) followed by `is_continuation:true` retries, and `message_end`
> events carrying `stop_reason:"error"` + a populated `error_message`.

## Risk

Scoped to the pi lapdog path (`pi_hooks.py` + `pi_lapdog_extension.ts`). Claude /
Codex hook paths are untouched. Continuation re-forwarding upserts by `span_id`, so
merged traces update cleanly.

### DIFF
--- a/tests/test_pi_hooks.py
+++ b/tests/test_pi_hooks.py
@@ -426,6 +426,53 @@ async def test_multiple_steps(agent):
     assert tools[0]["parent_id"] == step0["span_id"]
 
 
+async def test_stale_turn_end_does_not_finalize_next_step(agent):
+    """Out-of-order turn_end for the previous turn must not emit an empty step."""
+    sid = "pi-stale-turn-end"
+    tool_call_id = "tc-a"
+    content1 = [
+        {"type": "toolCall", "id": tool_call_id, "name": "read", "arguments": {"path": "foo.py"}},
+    ]
+    content2 = [{"type": "text", "text": "Done!"}]
+
+    await _post(agent, _session_start(sid))
+    await _post(agent, _agent_start(sid))
+    await _post(agent, _turn_start(sid, turn_index=0))
+    await _post(agent, _message_start(sid))
+    await _post(agent, _message_end(sid, content=content1, stop_reason="toolUse"))
+    await _post(agent, _tool_start(sid, tool_call_id, "read"))
+    await _post(agent, _tool_end(sid, tool_call_id, "read", "file contents"))
+
+    # Pi can emit the next turn_start before the previous turn_end. The stale
+    # turn_end(0) should not close turn 1 before its message_start/message_end.
+    await _post(agent, _turn_start(sid, turn_index=1))
+    await _post(agent, _turn_end(sid, turn_index=0))
+    await _post(agent, _message_start(sid))
+    await _post(agent, _message_end(sid, content=content2, stop_reason="stop"))
+    await _post(agent, _turn_end(sid, turn_index=1))
+    await _post(agent, _agent_end(sid))
+
+    resp = await agent.get("/claude/hooks/spans")
+    spans = _spans(await resp.json())
+    session_spans = [s for s in spans if s.get("session_id") == sid]
+
+    root = [s for s in session_spans if s["parent_id"] == "undefined"][0]
+    steps = sorted(_by_kind(session_spans, "step"), key=lambda s: s["name"])
+    llms = _by_kind(session_spans, "llm")
+    tools = _by_kind(session_spans, "tool")
+
+    assert [s["name"] for s in steps] == ["inference-0", "inference-1"]
+    assert [s["parent_id"] for s in steps] == [root["span_id"], root["span_id"]]
+    assert len(llms) == 2
+    assert len(tools) == 1
+
+    step1 = steps[1]
+    assert step1["meta"]["output"]["value"] == "Done!"
+    assert step1["meta"]["metadata"]["turn_index"] == 1
+    assert any(llm["parent_id"] == step1["span_id"] for llm in llms)
+    assert all(any(child.get("parent_id") == step["span_id"] for child in session_spans) for step in steps)
+
+
 async def test_trajectory_tags(agent):
     """Root has trajectory.semantic_type:turn, step has trajectory.semantic_type:agent_message."""
     sid = "pi-tags"
--- a/tests/test_pi_hooks.py
+++ b/tests/test_pi_hooks.py
@@ -51,6 +51,7 @@ def _agent_start(
     model_id="claude-sonnet-4-20250514",
     model_provider="anthropic",
     cwd=None,
+    is_continuation=False,
 ):
     event = {
         "session_id": session_id,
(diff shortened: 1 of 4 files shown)


tests/test_pi_hooks.py
```diff
@@ -775,3 +826,165 @@ async def test_thinking_block_sets_has_thinking(agent):
     assert step["meta"]["metadata"].get("has_thinking") is True
     # Thinking text should NOT appear in output (only text blocks)
     assert step["meta"]["output"]["value"] == "Here's my answer."
+
+
+# ------------------------------------------------------------------
+# Bug 1: LLM errors should produce an error LLM span (not an empty step)
+# ------------------------------------------------------------------
```
sabrenner: we don't need these kinds of comment blocks, the test names are descriptive enough for the regression case they're covering

sabrenner (approved):
qa'd and lgtm!

tests/test_pi_hooks.py
tomshen: Removed the redundant regression comment blocks in fc5553c.

PR #382, opened 2026-05-28
Lapdog: add session link status item to pi and Claude Code

tomshen:
## Summary

screenshot

Screenshot 2026-05-29 at 3 16 19 PM

Adds a clickable **🐶 lapdog** status item to both **pi** and **Claude Code**
that deep-links to the current session in the hosted lapdog dashboard
(`https://lapdog.datadoghq.com/?sessionId=<session id>`). The pi extension also
registers a `/lapdog` slash command for terminals without OSC-8 hyperlink
support.

### pi (`lapdog/pi_lapdog_extension.ts`)
- On `session_start`, sets a footer status entry (`ctx.ui.setStatus`) showing a
  dog emoji next to an accent-colored, OSC-8 hyperlinked `lapdog` label.
- Clears the status entry on `session_shutdown`.
- Registers a `/lapdog` command that prints the session URL and copies it to the
  clipboard — works in any terminal regardless of OSC-8 support.

### Claude Code
- `lapdog/claude_statusline.py` (new): reads Claude's stdin JSON, extracts
  `session_id`, and prints the same clickable `🐶 lapdog` OSC-8 hyperlink. Prints
  nothing when there is no active session.
- `lapdog/cli.py`: `lapdog claude` now wires the status line into
  `~/.claude/settings.json` (pointing at `python -m lapdog.claude_statusline`),
  and `lapdog uninstall` removes it. Claude renders only one status line, so an
  existing user `statusLine` is never overwritten — lapdog warns and skips. Only
  the lapdog entry is removed on uninstall.

Claude Code has no supported way for a *plugin* to contribute the main status
line (plugin `settings.json` only honors `agent`/`subagentStatusLine`, and
`plugin.json` has no `statusLine` field), so the settings.json wiring is the
standard path. The pip-installed module path is also more stable than the
churning marketplace-cache plugin path.

### Shared
- `lapdog/constants.py` (new): hosts `LAPDOG_DASHBOARD_URL` +
  `session_dashboard_url()`, now used by `cli.py`, `claude_statusline.py`, and
  `lapdog_ascii_art.py`. The pi extension keeps its own TS copy since it can't
  import Python.

## How to use

Open the link with **⌘+click** on macOS (**Ctrl+click** on Linux) — a plain
click is intentionally a no-op in most terminals.

## Test plan

- pi: `tsc` typecheck against pi's bundled types — passes. Live PTY smoke test of
  `pi -e <ext>` confirmed the footer renders `🐶 lapdog` with an intact OSC-8
  hyperlink carrying a real `sessionId`. Manually verified ⌘+click opens the
  session in Ghostty.
- Claude status line: `black` / `flake8` / `mypy` clean. Unit-tested the
  settings.json install/remove logic against a temp HOME — fresh install,
  idempotent re-run, clean removal, preserves an existing non-lapdog
  `statusLine` (and other keys), and never removes a user's own `statusLine`.
- `echo '{"session_id":"abc 123"}' | python -m lapdog.claude_statusline` emits a
  correctly URL-encoded OSC-8 hyperlink; empty session prints nothing.

## Out of scope

- A `/lapdog` slash command for Claude Code (Claude slash commands can't run
  bash-and-exit without invoking the model).

### DIFF
--- a/lapdog/claude_statusline.py
+++ b/lapdog/claude_statusline.py
@@ -1,0 +1,51 @@
+"""Claude Code status line script for lapdog.
+
+Claude Code invokes a configured ``statusLine`` command for every render,
+passing a JSON blob on stdin that includes the current ``session_id`` (see
+https://code.claude.com/docs/en/statusline). We read it and print a single
+clickable ``🐶 lapdog`` item that deep-links to this session in the hosted
+lapdog dashboard via an OSC-8 terminal hyperlink.
+
+Wired up by ``lapdog claude`` (see ``cli.py``); run standalone as:
+
+    echo '{"session_id": "abc"}' | python -m lapdog.claude_statusline
+"""
+
+import json
+import sys
+
+from lapdog.constants import session_dashboard_url
+
+# Dog face shown next to "lapdog", matching the pi extension footer status.
+DOG_EMOJI = "🐶"
+# ANSI styling for the label (purple accent + reset), mirroring the pi
+# extension's accent-colored "lapdog" label.
+_ACCENT = "\033[38;5;141m"
+_RESET = "\033[0m"
+
+
+def _hyperlink(text: str, url: str) -> str:
+    """Wrap text in an OSC-8 terminal hyperlink."""
+    return f"\033]8;;{url}\033\\{text}\033]8;;\033\\"
+
+
+def render(session_id: str) -> str:
+    label = f"{_ACCENT}lapdog{_RESET}"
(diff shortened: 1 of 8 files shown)


lapdog/claude_statusline.py
```diff
@@ -0,0 +1,51 @@
+"""Claude Code status line script for lapdog.
+
+Claude Code invokes a configured ``statusLine`` command for every render,
```
tomshen, line 3: I'm curious how we feel about this behavior

PR #384, opened 2026-05-29
fix(lapdog): drop duplicate Claude Code hook deliveries that emit phantom traces

tomshen:
## Problem

lapdog emits **double traces for each Claude Code user message — the first errored with a wrong (≈0) duration**.

Root cause: the lapdog hooks get registered in **two places at once** — a stale `~/.claude/settings.json` block left over from the pre-plugin `--enable-claude-code-hooks` flow (removed in #353/#356) **plus** the `lapdog` Claude Code plugin. Both run the same `curl`, so every hook event is POSTed to `/claude/hooks` twice.

The duplicate `UserPromptSubmit` is what produces the symptom:
1. 1st `UserPromptSubmit` → starts a fresh trace, creates root A (`root_span_emitted=False`).
2. 2nd (duplicate) `UserPromptSubmit` → sees `root_span_emitted is False` + an existing `_root_span_ref`, assumes the turn was interrupted, calls `_finalize_interrupted_turn(A)` → stamps A `status="error"`, `"interrupted by user"`, `duration = now - start_ns` (a few µs) → then starts *another* fresh trace, root B.

So every turn gets a phantom errored zero-duration root plus the real one. The state machine is correct under single delivery — verified that 2 clean turns → 2 clean roots; a single duplicated `UserPromptSubmit` → 1 errored + 1 ok root.

## Fix

**B — Harden the endpoint against duplicate delivery.** `handle_hook` now drops byte-identical hook bodies seen within a short window (`_DUPLICATE_EVENT_WINDOW_NS`, 5s) before they mutate session state. The check-and-record is synchronous so it's atomic across concurrently-handled requests, and it runs *before* the `Stop` sleep so duplicate `Stop`s are dropped too. `SubagentStart`/`SubagentStop` are **excluded** — distinct concurrent subagents legitimately share an identical payload (locked in by `test_concurrent_subagents_parent_correctly`). Duplicates are still recorded in `/claude/hooks/raw` for debugging; only dispatch/assembly is suppressed.

**C — Migrate users off the legacy registration.** `lapdog claude` (when auto-installing the plugin) and `lapdog uninstall` now strip leftover `/claude/hooks` entries from Claude Code's `settings.json` (honoring `CLAUDE_CONFIG_DIR`). Surgical: only lapdog commands are removed, unrelated hooks in shared blocks are preserved, and the file is left byte-for-byte unchanged when there's nothing to remove.

## Tests

- `test_duplicate_user_prompt_submit_emits_single_root` — the reported bug; one root, `ok`, no error.
- `test_duplicate_tool_use_emits_single_tool_span` — duplicate Pre/PostToolUse collapse to one span.
- `test_distinct_prompts_not_deduplicated` — two real turns still produce two roots.
- `test_concurrent_subagents_parent_correctly` (existing) — guards the subagent exclusion.
- `test_remove_legacy_claude_code_hooks_*` — strips lapdog entries / keeps unrelated / no-op when clean / missing file.

All 78 tests in the affected suites pass; `black`/`flake8` clean; new code is `mypy`-clean (pre-existing `trace.py` errors are unrelated).

Claude session: `821d4e14-1e83-4266-b4cb-bf8c63c3dbb4`
Resume: `claude --resume 821d4e14-1e83-4266-b4cb-bf8c63c3dbb4`

### DIFF
--- a/ddapm_test_agent/claude_hooks.py
+++ b/ddapm_test_agent/claude_hooks.py
@@ -7,6 +7,7 @@ into LLMObs-format spans that can be queried through the Event Platform APIs.
 import asyncio
 import getpass
 import gzip
+import hashlib
 import json
 import logging
 import os
@@ -36,7 +37,6 @@ from .coding_agent_metadata import project_metadata_tags
 from .coding_agent_metadata import resolve_project_metadata
 from .llmobs_event_platform import with_cors
 
-
 log = logging.getLogger(__name__)
 
 _HOSTNAME = socket.gethostname()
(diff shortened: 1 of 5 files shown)


tomshen:
We mentioned this in a release note: https://github.com/DataDog/dd-apm-test-agent/releases/tag/v1.55.0

So let's skip shipping this fix unless it becomes a common issue.

PR #387, opened 2026-06-04
only set sampling mechanism if nonzero

hannahkm:


releasenotes/notes/fix-sampling-mechanism-representation-for-ETP-6277a54ab5597635.yaml
```diff
@@ -0,0 +1,11 @@
+---
+prelude: >
+    Replace this text with content to appear at the top of the section for this
```
ajgajg1134: looks like you need to remove this prelude section?

hannahkm: hehe, you're right. fixed!

ajgajg1134 (approved):

PR #390, opened 2026-06-05
fix(lapdog): nest backfilled Claude subagents instead of splitting sessions

tomshen:
## Context

Follow-up to #357. Review feedback ([discussion r3364874384](https://github.com/DataDog/dd-apm-test-agent/pull/357#discussion_r3364874384)): when a Claude session spawned a subagent (e.g. `Explore`), backfill **split it into two different sessions** rather than nesting it, which exploded the number of sessions.

## Root cause

Claude Code stores each subagent's conversation in a **separate file**:

```
~/.claude/projects/<encoded-cwd>/<session-id>/subagents/agent-<id>.jsonl
```

The backfill walker's `rglob("*.jsonl")` picked these up as independent sessions (each gets its own `session_id = agent-<id>`), so one logical session became many. The earlier `Task` tool_use/tool_result handling only created an agent *span* in the main chain — it never consumed the subagent's actual transcript.

## Fix

**Client (`lapdog/backfill_claude.py`)**
- `_iter_session_files` now excludes any path under a `subagents/` directory, so subagent transcripts are no longer walked as standalone sessions.
- `_backfill_one` gathers a session's subagent transcripts (`<session-id>/subagents/**.jsonl`) and bundles them into the POST under a new `subagents` key.

**Server (`ddapm_test_agent/claude_backfill.py`)**
- `session_to_spans(..., subagents=None)` indexes subagent transcripts by their launch prompt. A subagent's first user message is a plain string equal to the parent `Task` tool_use's `input.prompt` — that's the cross-file link (verified against real transcripts).
- When a `Task` (or any subagent-spawning) tool call is matched, the subagent's step/llm/tool spans are built **under that Task agent span, in the same trace and session** — mirroring the live `claude_hooks`/`claude_proxy` pipeline. Token usage/cost rolls up onto the Task agent span.
- Nested subagents recurse. Any transcript that can't be matched falls back to its own trace under the **same** `session_id` (never a separate session).
- Tool-name-agnostic: matches by prompt, so it works whether the spawning tool is named `Task` or `Agent`.

## Tests

Added coverage in `tests/test_backfill_claude.py`:
- subagent transcript nests under the Task agent span in one trace/session, with token roll-up
- nesting works when the spawning tool isn't named `Task`
- orphan (unmatched) subagent stays in the same session
- walker excludes `subagents/` files; `_backfill_one` bundles them into the parent payload

## QA

Ran the real backfill (`lapdog claude --backfill` path) against an actual session containing a real `Explore` subagent transcript:
- **Before:** 2 `.jsonl` files → 2 sessions.
- **After:** 1 session forwarded; the subagent appears as a nested `agent` span (`agent_id` recorded) in the **same trace** as its parent turn, with its full 6-step sub-conversation (haiku LLM + `Read` tool spans) underneath. Re-running is idempotent (`already_backfilled`).

### DIFF
--- a/ddapm_test_agent/claude_backfill.py
+++ b/ddapm_test_agent/claude_backfill.py
@@ -17,10 +17,15 @@ pipeline does, so backfilled sessions are indistinguishable in the UI:
     each UserPromptSubmit opens a new trace).
   * One LLM span per ``type=assistant`` entry, with model + token usage +
     cost computed via ``claude_cost_tracker.compute_cost_metrics``.
-  * One tool span per tool_use / tool_result pair, parented to the LLM span
-    that requested it. Claude Code ``Task`` tool uses become child agent
-    spans so backfilled subagents are visible as agents rather than generic
-    tools.
+  * One tool span per tool_use / tool_result pair, parented to the step span
+    for that inference cycle. Claude Code ``Task`` tool uses become child
+    agent spans parented to the same step span, so backfilled subagents are
+    visible as agents rather than generic tools while matching live traces.
+  * Subagent (e.g. ``Explore``) conversations — which Claude writes to
+    separate ``<session-id>/subagents/agent-*.jsonl`` files and the client
+    bundles into the ``subagents`` payload — are nested *under* the ``Task``
+    agent span that launched them (matched by launch prompt), in the same
+    trace and session, instead of being backfilled as standalone sessions.
 
 Timestamps come from the entries' ``timestamp`` field (ISO-8601). Durations
 between adjacent entries are used as best-effort span durations.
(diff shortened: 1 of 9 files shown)


ddapm_test_agent/llmobs_event_platform.py
```diff
@@ -527,10 +532,10 @@ def _build_trace_aggregates(
...
             "num_evaluations_failed": num_evaluations_failed,
             "number_of_errors": number_of_errors,
-            "input_tokens": sum(span.get("metrics", {}).get("input_tokens", 0) for span in trace_spans),
-            "output_tokens": sum(span.get("metrics", {}).get("output_tokens", 0) for span in trace_spans),
-            "total_tokens": sum(span.get("metrics", {}).get("total_tokens", 0) for span in trace_spans),
-            "estimated_total_cost": sum(span.get("metrics", {}).get("estimated_total_cost", 0) for span in trace_spans),
+            "input_tokens": sum(span.get("metrics", {}).get("input_tokens", 0) for span in llm_spans),
+            "output_tokens": sum(span.get("metrics", {}).get("output_tokens", 0) for span in llm_spans),
+            "total_tokens": sum(span.get("metrics", {}).get("total_tokens", 0) for span in llm_spans),
+            "estimated_total_cost": sum(span.get("metrics", {}).get("estimated_total_cost", 0) for span in llm_spans),
```
tomshen, line 538: The trace-level cost/token rollups need to sum only LLM spans because agent spans can carry rollup metrics copied from their child LLM calls. Backfilled Claude subagents now keep those metrics on the Task agent span for span-level inspection, so including all span kinds here would count the same usage once on the LLM span and again on the agent span.

tomshen (commented):
Adding context on the trace cost rollup change.

ddapm_test_agent/claude_backfill.py
```diff
@@ -419,15 +463,243 @@ def _finalize_turn(turn: Dict[str, Any]) -> None:
...
+    agent_span: Dict[str, Any],
+    trace_id: str,
+    entries: List[Dict[str, Any]],
+    subagent_index: Dict[str, List[Dict[str, Any]]],
+    *,
+    standalone: bool = False,
+) -> List[Dict[str, Any]]:
+    """Convert a subagent's transcript into spans nested under ``agent_span``.
+
+    Mirrors the per-turn logic in ``session_to_spans`` but uses ``agent_span``
```
sabrenner, line 577: without looking too closely, is there some logic we can refactor here?

sabrenner (approved):
code roughly lgtm! noticed one thing while QAing

i do see subagents now properly nested under their session, but it just looks a bit different from the original. here are a couple original shots of a session trace w a subagent:

Image

Image

however, when backfilled, the subagent is nested under a step span

Image

Image

the original bug is fixed, but the interpretations of the data are inconsistent (maybe subagents actually _should_ be under step spans, but either way it's different between "live" sessions and backfilled sessions)

approving though otherwise!

ddapm_test_agent/llmobs_event_platform.py
sabrenner: makes sense to me

ddapm_test_agent/claude_backfill.py
tomshen: Refactored this into `_finish_pending_tool_result(...)` so top-level and nested subagent transcript handling share the same tool_result path. That helper now owns the subagent claim/build/recurse flow and the step-duration update, so the two call sites stay consistent.

tomshen:
Parenting should be fixed now

Claude backfill:
Claude backfill subagent parenting
Claude live:
Claude live subagent parenting
Codex for comparison:
Codex subagent parenting

PR #393, opened 2026-06-10
fix(lapdog): do not crash python process when starting lapdog

sabrenner:
Starting lapdog with a coding agent would intermittently cause Python crashes. From investigation, it seems this happens on Python 3.13. This PR attempts to fix it by using `urllib.request` instead of `requests`. It's also possible forking processes here has some issues and subprocessing would work better, but just attempting this for now.

### DIFF
--- a/lapdog/cli.py
+++ b/lapdog/cli.py
@@ -15,9 +15,9 @@ from typing import Dict
 from typing import List
 from typing import Optional
 from typing import Tuple
+import urllib.error
+import urllib.request
 import uuid
-
-import requests
 
 from lapdog import backfill_claude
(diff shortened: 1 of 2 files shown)


Yun-Kim (approved):
Looks like a harmless change at worst, are we able to reproduce the issue and confirm this solves it?

sabrenner:
> are we able to reproduce the issue and confirm this solves it?

i had previously seen this issue so yes, but haven't recently. this is more of a best-effort attempt. i'll actually try and get it to surface one more time before merging to confirm. if i can't though i'm just gonna send it

PR #401, opened 2026-07-20
fix(lapdog): route path-form launcher invocations to dedicated launcher

ncybul:
## Summary

Fixes [MLOB-7870](https://datadoghq.atlassian.net/browse/MLOB-7870).

lapdog dispatches on an exact string match of the first argument, so only the bare names `claude`, `pi`, and `codex` reach their dedicated launchers. Invoking a managed launcher by an explicit **path** (e.g. `lapdog ~/.local/bin/claude`) — the same binary `which claude` resolves to — doesn't match, so it falls through to the generic `cmd_exec` wrapper instead.

`cmd_exec` injects a `PYTHONPATH` pointing at `lapdog/bootstrap`, which is then inherited by every Python subprocess spawned for the rest of the session (hooks, Bash tool calls, MCP servers). Each fails `sitecustomize`'s `import ddtrace` check and exits with `[lapdog] ddtrace is not installed`. The result: two invocations of the *identical* binary behave differently — the bare name works, the full path silently breaks the session.

## Fix

- Add `_canonical_launcher(target)`: maps an invocation target to a managed launcher name (`claude`/`pi`/`codex`) whether it's spelled as a bare name or an explicit path.
- `main()` resolves the target before dispatch, so path-form invocations of any managed launcher route to the dedicated launcher (no `PYTHONPATH` injection). Unknown targets still fall through to `cmd_exec` unchanged.

Matching is **resolve-and-compare** (`realpath(target) == realpath(which(launcher))`), not basename. The launchers install as versioned binaries fronted by a symlink (e.g. Claude Code is `~/.local/share/claude/versions/<ver>` behind a `claude` symlink), so a basename check would see `<ver>`, not `claude`. Comparing resolved paths handles this and matches the ticket's acceptance criteria ("any path resolving to the same binary").

## Tests

6 new tests in `tests/test_lapdog_cli.py` covering bare name, path-via-symlink resolution, unknown targets, a launcher-named file that does *not* resolve to the managed binary, and `main()` routing (both the reroute and the unchanged `cmd_exec` fall-through). Full `test_lapdog_cli.py` suite passes (54 tests).

Claude session: `f2ab1d44-6581-4b2c-95e5-f546fab5abfa`
Resume: `claude --resume f2ab1d44-6581-4b2c-95e5-f546fab5abfa`

[MLOB-7870]: https://datadoghq.atlassian.net/browse/MLOB-7870?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lapdog/cli.py
+++ b/lapdog/cli.py
@@ -31,6 +31,11 @@ from lapdog.paths import LOG_FILE
 from lapdog.paths import PID_FILE
 
 LAPDOG_COMMANDS = ["start", "stop", "status", "claude", "pi", "codex", "uninstall"]
+# Managed launchers that also exist as external binaries a user might invoke by
+# an explicit path (e.g. ``lapdog ~/.local/bin/claude``). Unlike start/stop/
+# status/uninstall, these must be recognized whether spelled as a bare name or
+# a path so they route to their dedicated launcher instead of ``cmd_exec``.
+_PATH_ROUTABLE_LAUNCHERS = ("claude", "pi", "codex")
 LAPDOG_USAGE = (
     "Usage: lapdog [OPTIONS] <command> [command-args...]\n"
     "Options must appear before <command>. Arguments after <command> are forwarded.\n"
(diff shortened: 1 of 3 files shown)


lapdog/cli.py
```diff
@@ -31,6 +31,11 @@
 from lapdog.paths import PID_FILE
 
 LAPDOG_COMMANDS = ["start", "stop", "status", "claude", "pi", "codex", "uninstall"]
+# Managed launchers that also exist as external binaries a user might invoke by
+# an explicit path (e.g. ``lapdog ~/.local/bin/claude``). Unlike start/stop/
+# status/uninstall, these must be recognized whether spelled as a bare name or
+# a path so they route to their dedicated launcher instead of ``cmd_exec``.
```
ncybul:
```suggestion
# Managed launchers that also exist as external binaries a user might invoke by
# an explicit path (e.g. ``lapdog ~/.local/bin/claude``)
```

lapdog/cli.py
```diff
@@ -1030,6 +1035,41 @@ def _consume_backfill_arg(args: List[str]) -> Tuple[List[str], bool]:
...
+    "[lapdog] ddtrace is not installed" (MLOB-7870).
+
+    A bare, unknown word (e.g. ``python``) returns None so it still falls
+    through to ``cmd_exec`` unchanged.
+
+    Matching is by resolved target, not basename: Claude Code installs as a
+    versioned binary (``.../versions/2.1.215``) with a ``claude`` symlink, so
+    ``realpath`` of the invoked path and of ``which claude`` both point at the
+    same versioned file even though its basename is not ``claude``.
+    """
```
ncybul, line 1046:
```suggestion
    """Map an invocation target to a managed launcher name, or return None.

    A managed launcher may be invoked either by bare name (``claude``) or by an
    explicit path that resolves to the same binary (``~/.local/bin/claude``,
    i.e. what ``which claude`` returns). Both must route to the dedicated
    launcher (``cmd_claude``/``cmd_pi``/``cmd_codex``). 

    A bare, unknown word (e.g. ``python``) returns None so it still falls
    through to ``cmd_exec`` unchanged.
    """
```

lapdog/cli.py
```diff
@@ -1030,6 +1035,41 @@ def _consume_backfill_arg(args: List[str]) -> Tuple[List[str], bool]:
...
+    if target.lower() in _PATH_ROUTABLE_LAUNCHERS:
+        return target.lower()
+
+    # Only treat the target as a path when it looks like one.
```
ncybul:
```suggestion
```

lapdog/cli.py
```diff
@@ -1052,10 +1092,14 @@ def main() -> None:
     lapdog_args, remaining = _parse_command(args[1:])
     lapdog_parsed_args = _parse_lapdog_args(lapdog_args)
 
-    sub_cmd = remaining[0].lower()
+    # Resolve path-form invocations of a managed launcher (e.g.
+    # ``lapdog ~/.local/bin/claude``) to their canonical command so they route
+    # to the dedicated launcher rather than the generic cmd_exec wrapper
+    # (MLOB-7870). Unknown targets fall through to cmd_exec unchanged.
```
ncybul:
```suggestion
```

sabrenner (approved):
lgtm!

PR #402, opened 2026-07-20
fix(trace): project v1 boolean span attributes as meta strings

AlexeyKuznetsov-DD:
# What Does This Do

Projects typed `v1` boolean span attributes into the legacy `meta` map as the strings `"true"`/`"false"` instead of converting them to numeric `metrics` (`1`/`0`).

# Motivation

Datadog's convention stores boolean tags in `meta` as `"true"`/`"false"` strings. The previous `metrics[key] = 1/0` behavior made booleans indistinguishable from real numeric metrics, breaking snapshot comparisons and tag-based filtering in system tests.

# Additional Notes

- Only affects the span-level `_convert_v1_attributes` projection; the span *event* attribute path (typed `bool_value`) is unchanged.
- Adds `tests/test_trace_v1_attributes.py` as a regression guard.

### DIFF
--- a/ddapm_test_agent/trace.py
+++ b/ddapm_test_agent/trace.py
@@ -1144,8 +1144,8 @@ def _convert_v1_attributes(
         if value_type == V1AnyValueKeys.STRING:
             meta[key] = _get_and_add_string(string_table, value)
         elif value_type == V1AnyValueKeys.BOOL:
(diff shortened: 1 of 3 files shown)


AlexeyKuznetsov-DD:
Decided to fix in `system-tests` instead.

PR #403, opened 2026-07-20
Allow Claude to add custom tags to its own Lapdog spans

yshapiro-57:
Background
----------------------
The intent of this PR is to create a way to integrate Lapdog with skill-based auto-experiments. The way that skill-based auto experiments work, we basically write a skill that says:

"Look around the codebase, try to create your own eval metric, make one improvement to the codebase, and compare eval metrics values before and after the improvement. Repeat five times. After each iteration, submit the eval metric score back to Datadog using its MCP server".

This allows the users to run the experiment without having to share their code with Datadog. Our UI for auto experiments is built on top of the eval metrics events that Claude will submit with this skill.

If the user happens to also have Lapdog installed, then we want to show the Lapdog traces in a separate tab in the auto experiment UI. This means we need some way to tag those traces with a specific auto experiment ID.

Changes
----------------------
My initial idea was to store the experiment ID in an environment variable. But that's a pain in the neck for several reasons:
- The Lapdog agent is a separate process that does not share environment with Claude
- Claude has no way of updating its own environment. It can only control the environment of the terminal shells that it creates
- By the time we tell Claude to set the environment variable and get around to it, the turn has already started and the initial root span has been created

So after conferring with Codex, I am going the route of adding a new `tags set` command to the Lapdog CLI:

```
lapdog tags set dd_auto_experiment_id:b5f657b1-7ed8-4bb7-92d4-f92e790c997c
```

Lapdog will internally store a mapping of session IDs to tags. The tags for a session ID will apply to all spans for that session that have not already been emitted by the time the tag is set, including the root span.

The only possible problem is that, over time, the mapping of sessions to tags may grow. But since I expect the cardinality of tags per session to be very low (for auto experiment, we just need the experiment_id and iteration), I do not think this will cause the agent to run out of memory.

Automated Tests
----------------------
```
cd /Users/$USER/go/src/github.com/DataDog/dd-apm-test-agent

python3.12 -m venv .venv
source .venv/bin/activate
pip install -e ".[testing]"

python -m pytest tests/test_claude_hooks.py tests/test_claude_steps.py tests/test_lapdog_cli.py tests/test_codex_hooks.py tests/test_pi_hooks.py
```

Manual Verification
-----------------------
To run a version of Lapdog that includes these changes, I used the following commands
```
cd /Users/$USER/go/src/github.com/DataDog/dd-apm-test-agent

python3.12 -m venv .venv
source .venv/bin/activate
python -m pip install -e .

which lapdog
# The above should output a path that ends with dd-apm-test-agent/.venv/bin/lapdog, not one in /opt
lapdog stop
lapdog start

lapdog claude --dangerously-skip-permissions
```

I then pasted the following prompts into Claude, one and then the other:

I want you to check if you have access to the lapdog CLI (for example, by running the command `which lapdog`). If you do have access to the CLI, please use it to add a new tag to the Lapdog telemetry with the following command: `lapdog tags set payload:test`. After adding that tag to the telemetry, check the current status of the following PR and tell me what it is: `https://github.com/DataDog/dd-apm-test-agent/pull/403`

Now I want you to update the telemetry tag by running `lapdog tags set payload:shmoduction` and, after doing so, pull up the description of the same PR, `https://github.com/DataDog/dd-apm-test-agent/pull/403`, and summarize for me what it does

I confirmed that for both of them, all spans get tagged with the `payload` tag, [which becomes searchable in our UI](https://app.datadoghq.com/llm/traces?query=%40ml_app%3Aclaude-code%20%40event_type%3Aspan%20payload%3Ashmoduction%20%40is_root_span%3Atrue&agg_m=count&agg_m_source=base&agg_t=count&fromUser=true&is_llm_session=false&refresh_mode=paused&selectedTab=overview&spanId=11521154653168241800&start=1784645069374&end=1784648669374&paused=true). For the first trace, I did not yet have my Lapdog forward spans to our backend, so only the second trace is visible in production. I inspected the first one on the Lapdog that was running on my laptop, and confirmed that it also has the `payload` tags on each span.

I repeated the same test for the Codex CLI as well and confirmed that [the tags also become attached to spans](https://app.datadoghq.com/llm/traces?query=%40ml_app%3Acodex%20%40event_type%3Aspan%20payload%3Atry_codex%20%40is_root_span%3Atrue&agg_m=count&agg_m_source=base&agg_t=count&fromUser=true&is_llm_session=false&refresh_mode=paused&selectedTab=overview&sidepanelContextScope=14218801265911369967&sidepanelContextScopeKind=agent&sp=%5B%7B%22p%22%3A%7B%22eventId%22%3A%22AwAAAZ-FdEwgeA2ykQAAABhBWi1GZEV3Z0FBRFpVUExqd0R0NkFBQUEAAAAkZjE5Zjg1NzUtNDczNC00NTk3LTg3MmItODliNmViMGUzY2NlAAATzA%22%7D%2C%22i%22%3A%22llm-obs-panel%22%7D%5D&spanId=14218801265911369967&start=1784649595772&end=1784650495772&paused=true) as intended.

For the latest version of the PR I reran the same test steps for Pi as well.

### DIFF
--- a/ddapm_test_agent/claude_hooks.py
+++ b/ddapm_test_agent/claude_hooks.py
@@ -308,7 +308,7 @@ class ClaudeHooksAPI:
     def __init__(self, link_tracker: Optional[ClaudeLinkTracker] = None) -> None:
         self._sessions: Dict[str, SessionState] = {}
         self._session_ids_by_token: Dict[str, Set[str]] = {}
-        self._tags_by_token: Dict[str, Dict[str, str]] = {}
+        self._pending_tags_by_token: Dict[str, Dict[str, str]] = {}
         self._assembled_spans: List[Dict[str, Any]] = []
         self._raw_events: List[Dict[str, Any]] = []
         self._link_tracker = link_tracker
@@ -341,7 +341,7 @@ class ClaudeHooksAPI:
 
     def _register_session_token(self, session_token: str, session_id: str) -> None:
         self._session_ids_by_token.setdefault(session_token, set()).add(session_id)
-        pending_tags = self._tags_by_token.get(session_token)
+        pending_tags = self._pending_tags_by_token.pop(session_token, None)
         if pending_tags:
             self._set_session_tags(self._get_or_create_session(session_id), pending_tags)
 
@@ -1902,19 +1902,42 @@ class ClaudeHooksAPI:
                 return web.json_response({"error": "tag keys and values must be non-empty strings"}, status=400)
             normalized[key.strip()] = value.strip()
 
-        token_tags = self._tags_by_token.setdefault(session_token, {})
-        token_tags.update(normalized)
-        session_ids = sorted(self._session_ids_by_token.get(session_token, set()))
-        for session_id in session_ids:
-            session = self._sessions.get(session_id)
-            if session is not None:
+        requested_session_id = body.get("session_id")
+        if requested_session_id is not None and (
+            not isinstance(requested_session_id, str) or not requested_session_id.strip()
+        ):
+            return web.json_response({"error": "session_id must be a non-empty string"}, status=400)
+
+        if isinstance(requested_session_id, str):
+            session_id = requested_session_id.strip()
+            session = self._get_or_create_session(session_id)
+            self._set_session_tags(session, normalized)
+            session_ids = [session_id]
+            applied_tags = dict(session.custom_tags)
+        else:
+            session_ids = sorted(self._session_ids_by_token.get(session_token, set()))
+            if len(session_ids) > 1:
+                return web.json_response(
+                    {
+                        "error": "launch token matches multiple sessions; provide session_id",
+                        "session_ids": session_ids,
+                    },
+                    status=409,
+                )
+            if session_ids:
+                session = self._get_or_create_session(session_ids[0])
                 self._set_session_tags(session, normalized)
+                applied_tags = dict(session.custom_tags)
+            else:
+                pending_tags = self._pending_tags_by_token.setdefault(session_token, {})
+                pending_tags.update(normalized)
+                applied_tags = dict(pending_tags)
         return web.json_response(
             {
                 "status": "ok",
                 "session_id": session_ids[-1] if session_ids else "",
                 "session_ids": session_ids,
-                "tags": token_tags,
+                "tags": applied_tags,
             }
         )
 
--- a/ddapm_test_agent/claude_hooks.py
+++ b/ddapm_test_agent/claude_hooks.py
@@ -345,14 +345,21 @@ class ClaudeHooksAPI:
         if pending_tags:
             self._set_session_tags(self._get_or_create_session(session_id), pending_tags)
 
-    def _append_span(self, span: Dict[str, Any]) -> None:
-        """Store a span, applying tags configured for its Claude session."""
+    def _apply_registered_session_tags(self, span: Dict[str, Any]) -> None:
         session_id = span.get("session_id")
-        if isinstance(session_id, str):
-            session = self._sessions.get(session_id)
-            if session is not None:
+        if not isinstance(session_id, str):
+            return
+        for raw_session_id, session in self._sessions.items():
+            if raw_session_id == session_id or session.session_id == session_id:
                 self._apply_session_tags(span, session)
-        self._assembled_spans.append(span)
+
+    def _append_span(self, span: Dict[str, Any], index: Optional[int] = None) -> None:
+        """Store a span, applying tags configured for its coding-agent session."""
+        self._apply_registered_session_tags(span)
+        if index is None:
+            self._assembled_spans.append(span)
+        else:
+            self._assembled_spans.insert(index, span)
 
     def _get_or_create_session(self, session_id: str) -> SessionState:
         """Get existing session or create a new one."""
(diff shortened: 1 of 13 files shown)


lapdog/claude_intercept.mjs
```diff
@@ -10,13 +10,17 @@
  * Environment variables:
  *   DDAPM_GATEWAY_URL       - Gateway URL (default: http://localhost:8126/claude/proxy)
  *   DDAPM_INTERCEPT_DEBUG   - "true" to enable stderr log messages
+ *   LAPDOG_SESSION_TOKEN    - Correlates hook events with this instrumented Claude process
```
yshapiro-57, line 13: I asked Codex why it needed to touch this file and what the new code does, and it came back with the following answer:
```
It connects the generated launch token to Claude’s actual session:

Injects lapdog_session_token into Claude hook payloads.
Lets the server associate lapdog tags set with the correct session.
Rewrites the installed plugin’s hardcoded hook URL when using a non-default port.
```

lapdog/cli.py
```diff
@@ -468,10 +479,74 @@ def cmd_claude(
...
+        if not separator or not key or not value:
+            print(f"[lapdog] Invalid tag {raw_tag!r}; expected key:value.", file=sys.stderr)
+            sys.exit(1)
+        tags[key] = value
+
+    session_token = os.environ.get("LAPDOG_SESSION_TOKEN", "")
+    lapdog_url = os.environ.get("LAPDOG_URL", "")
+    if not session_token or not lapdog_url:
+        print(
+            "[lapdog] No instrumented Claude session found. Run this command from inside 'lapdog claude'.",
```
sabrenner:
```suggestion
            "[lapdog] No instrumented Claude session found. Run this command from inside a Claude session started with 'lapdog claude'.",
```
just to make it a little more clearer. should we also mention codex here too, since we support that / this should work for codex?

lapdog/cli.py
```diff
@@ -468,10 +479,74 @@ def cmd_claude(
...
+    except (OSError, ValueError) as exc:
+        print(f"[lapdog] Failed to set session tags: {exc}", file=sys.stderr)
+        sys.exit(1)
+
+    formatted_tags = ", ".join(f"{key}:{value}" for key, value in tags.items())
+    session_id = result.get("session_id")
+    if session_id:
+        print(f"[lapdog] Tagged Claude session {session_id}: {formatted_tags}")
+    else:
+        print(f"[lapdog] Queued tags for the current Claude session: {formatted_tags}")
```
sabrenner: we should probably make this provider-agnostic if we support both claude, codex, maybe pi, etc. for this command

sabrenner (commented):
code lgtm! i followed up with one possible shortfall offline, but just leaving one comment on wording here.

should we also support Pi as well? or just claude/codex intentionally?

yshapiro-57:
I have added support for Pi, and also confirmed that if I run two Codex sessions in parallel with the latest version of the code, then tags applied in each of them modify only that session's spans and not any other session's. This PR should be ready for a new look.

ddapm_test_agent/claude_hooks.py
```diff
@@ -1826,6 +1946,81 @@ async def handle_spans(self, request: Request) -> web.Response:
...
+        tags = body.get("tags")
+        if not isinstance(tags, dict) or not tags:
+            return web.json_response({"error": "tags must be a non-empty object"}, status=400)
+
+        normalized: Dict[str, str] = {}
+        for key, value in tags.items():
+            if not isinstance(key, str) or not key.strip() or not isinstance(value, str) or not value.strip():
+                return web.json_response({"error": "tag keys and values must be non-empty strings"}, status=400)
+            normalized[key.strip()] = value.strip()
+        reserved_keys = sorted(set(normalized).intersection(_RESERVED_SESSION_TAG_KEYS))
```
sabrenner, line 1970: do we need the `sorted` here if we're just checking below if the `reserved_keys` has any entries?

lapdog/cli.py
```diff
@@ -468,10 +481,84 @@ def cmd_claude(
...
+            sys.exit(1)
+        tags[key] = value
+
+    session_token = os.environ.get("LAPDOG_SESSION_TOKEN", "")
+    lapdog_url = os.environ.get("LAPDOG_URL", "")
+    target_session_id = os.environ.get("CODEX_THREAD_ID") or os.environ.get("PI_SESSION_ID") or None
+    if not session_token or not lapdog_url:
+        print(
+            "[lapdog] No instrumented coding-agent session found. "
+            "Run this command from inside a session started with 'lapdog claude', 'lapdog codex', or 'lapdog pi'.",
```
sabrenner: might be nice to make this log line dynamic wrt `_PATH_ROUTABLE_LAUNCHERS` so that if we add more support down the line, they show up here automatically without needing to remember to update this log line

sabrenner (approved):
two non-blocking thoughts but lgtm! 

i tested locally with Claude, Codex, and Pi, and ran two concurrent sessions in each. Manually instructing each to run `lapdog tags set sam:test-{num}-{coding_agent}`, ie `sam:test-1-claude`, to make sure the correct tags ended up on the correct session's spans, worked great!

ddapm_test_agent/claude_hooks.py
yshapiro-57: Codex says it is trying to make the error response deterministic, and `reserved_keys` is included in the error response on line 1973.

I do not have a strong opinion personally, I think it's not a big deal if they go in a random order, let me know if you would like me to change it.

PR #404, opened 2026-07-24
Break Lapdog payloads into chunks of 20 spans each

yshapiro-57:
## Summary
When testing auto experiments, which produce traces that are often 30 minutes long and contain hundreds of spans, I noticed that sometimes the trace would show up in my local Lapdog UI but never make it to us5.datadoghq.com, even though Lapdog logs showed the payload submission request getting a 200. I assume these large traces are running into some sort of payload size restriction either in our own intake or in EvP, and that prevents them from coming through.

In this vibecoded PR I am making Lapdog split up payloads into batches of at most 20 spans each before forwarding them to the Datadog backend. With these changes I was able to see my large traces show up both in the local Lapdog UI and on the Datadog backend.

## Test plan
- [x] `python -m pytest tests/test_claude_hooks.py tests/test_codex_hooks.py tests/test_pi_hooks.py -q`
- [x] verify 20, 21, and 45-span payload boundaries

I was also able to run an auto experiment and get a 24-minute-long trace with lots of spans to show up in the Datadog UI after I tested with these changes in place.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/ddapm_test_agent/claude_hooks.py
+++ b/ddapm_test_agent/claude_hooks.py
@@ -104,14 +104,7 @@ def _to_json_str(value: Any) -> str:
 class PendingToolSpan:
     """Tracks a tool invocation between PreToolUse and PostToolUse."""
 
-    def __init__(
-        self,
-        span_id: str,
-        tool_name: str,
-        tool_input: Any,
-        parent_id: str,
-        start_ns: int,
-    ) -> None:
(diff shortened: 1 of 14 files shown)


sabrenner (approved):
change looks clean, and when i ran a claude session locally and had it do operations that generated a bunch of spans, all the spans showed up in lapdog and in the main app. also further shown by the log line
```
INFO:ddapm_test_agent.claude_hooks:Splitting 87 spans into 5 backend requests of at most 20 spans
```

PR #416, opened 2026-09-09
feat(lapdog): add cost tracking for fable-5.1 and astra-6

sabrenner:
Adds cost tracking for the `astra-6` and `fable-5.1` models from OpenAI and Anthropic respectively for `lapdog`.

https://developers.openai.com/api/docs/pricing
https://claude.com/pricing#api

### DIFF
--- a/ddapm_test_agent/claude_cost_tracker.py
+++ b/ddapm_test_agent/claude_cost_tracker.py
@@ -11,7 +11,7 @@ and the metric keys expected by the web-ui LLM observability span detail view:
     estimated_cache_write_input_cost
     estimated_cache_read_input_cost
 
-Pricing data last updated 2026-08 based on:
+Pricing data last updated 2026-09 based on:
   * Anthropic public pricing: https://www.anthropic.com/pricing#api
   * pi-ai models.generated.js (the data Claude Code itself uses to compute
     the self-reported per-turn cost shown in its UI):
@@ -24,7 +24,9 @@ whereas Opus 4 / 4.1 / 3 still cost $15 / $75 / $1.50 / $18.75 per Mtok.
 Using the old Opus rates for 4.5+ overcharges by exactly 3x.
 
 The Claude 5 family adds Sonnet 5 ($3 / $15 per Mtok, same tier as Sonnet 4.6)
-and Fable 5 / Mythos 5 ($10 / $50 per Mtok — the most capable tier).  Sonnet 5
+and Fable 5 / Mythos 5 ($10 / $50 per Mtok — the most capable tier). Fable 5.1
+keeps those input and output rates while reducing cache reads to $0.25 per Mtok.
+Sonnet 5
 carries an introductory rate of $2 / $10 per Mtok through 2026-08-31; because
 these values are only estimates, the table uses the higher ongoing list price
 ($3 / $15) as a conservative upper bound rather than the temporary intro rate.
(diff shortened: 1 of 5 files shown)


ddapm_test_agent/claude_cost_tracker.py
```diff
@@ -87,6 +89,14 @@ class _PriceTier:
 _ONE_TIER = 0  # sentinel: no upper bound on a single tier
 
 _PRICING: List[Tuple[str, List[_PriceTier]]] = [
+    # ---- Fable 5.1 ---------------------------------------------------------
+    # $10 / $50 / $0.25 cache-read / $12.50 cache-write per Mtok.
+    # This must precede Fable 5 so prefix matching does not use the older
+    # model's $1.00 cache-read rate.
+    (
+        "claude-fable-5-1",
```
tomshen, line 97: [P2] Anthropic's Fable 5.1 docs say `claude-mythos-5-1` shares Fable 5.1's pricing: https://platform.claude.com/docs/en/models/fable-5-1/overview. With the current ordering, that ID falls through to the existing `claude-mythos-5` prefix, so cache reads are charged at $1/MTok instead of $0.25/MTok. In manual gateway QA, 500 cached tokens produced 500,000 nanodollars instead of 125,000. Please add the 5.1 Mythos prefix before `claude-mythos-5` and cover it alongside Fable 5.1.

ddapm_test_agent/codex_cost_tracker.py
```diff
@@ -29,12 +31,32 @@ class _OpenAIPrice:
...
+    _OpenAIPrice(
+        "gpt-5.6-terra",
+        input_price=2_000,
+        cached_input=200,
+        output=12_000,
+        long_context_threshold=272_000,
+    ),
+    _OpenAIPrice(
+        "gpt-5.6",
+        input_price=5_000,
```
tomshen: [P2] OpenAI's current GPT-5.6 Sol page lists $4 input / $0.40 cached / $20 output per MTok and says that promotional pricing lasts at least through Nov 21, 2026: https://developers.openai.com/api/docs/models/gpt-5.6-sol. This row still uses $5 / $0.50 / $30, so the estimates—including the new long-context path—are 25% high for input/cache and 50% high for output. If the higher post-promo estimate is intentional, could we document that explicitly as the Claude tracker does; otherwise the row and tests should use the current rates.

tomshen (commented):
Manual QA covered exact and date-suffixed Astra IDs, the 272,000/272,001 long-context boundary, and the Codex-proxy and Lapdog/PI gateway call paths. Astra and Fable 5.1 behave as expected; I found the two cost-accuracy gaps noted inline.

ddapm_test_agent/codex_cost_tracker.py
sabrenner: yeah we can hardcode the promotional pricing for now and change it in november

ddapm_test_agent/claude_cost_tracker.py
sabrenner: ah missed that, thanks!

tomshen (approved):
Re-reviewed the follow-up commit and repeated manual QA through the Codex proxy and Lapdog/PI gateway paths. The Mythos 5.1 prefix and GPT-5.6 Sol promotional pricing findings are fixed, boundary behavior remains correct, legacy pricing is preserved, and CI is green.

PR #417, opened 2026-09-09
refactor(lapdog): pull out lapdog into its own directory by extending the testagent server

sabrenner:
Refactors the `lapdog` &harr; `ddapm-test-agent` relationship so `ddapm-test-agent` does not have to know about lapdog or its implementation. This makes it a bit easier to add on routes for lapdog-specific things without polluting the test agent more.

### DIFF
--- a/lapdog/README.md
+++ b/lapdog/README.md
@@ -92,7 +92,7 @@ docker run --rm \
     -p 4318:4318 \
     -p 4317:4317 \
     ghcr.io/datadog/dd-apm-test-agent/ddapm-test-agent:latest \
-    ddapm-test-agent --lapdog-mode
+    python -m lapdog.server
 ```
 
 Then point your application at the host: `DD_TRACE_AGENT_URL=http://localhost:8126`.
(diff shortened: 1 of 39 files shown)


lapdog/README.md
```diff
@@ -92,7 +92,7 @@ docker run --rm \
     -p 4318:4318 \
     -p 4317:4317 \
     ghcr.io/datadog/dd-apm-test-agent/ddapm-test-agent:latest \
-    ddapm-test-agent --lapdog-mode
+    python -m lapdog.server
```
sabrenner, line 95: hmm from this change, the whole refactor introduces a breaking change for Docker users. although as most likely use lapdog via CLI, it would not be a breaking change, but nevertheless...

tomshen (approved):
Reviewed the route ownership refactor at 3e1e710. No actionable issues found.

QA performed:
- Launched the base test agent and Lapdog server from an isolated worktree at the PR head.
- Verified /info returns 200 for both servers.
- Verified /claude/hooks/spans is absent from the base agent (404) and present on the Lapdog server (200).
- Built and installed the wheel, then verified the packaged ddapm-test-agent and lapdog console entry points plus python -m lapdog.server.
- Confirmed the updated Docker path is documented as python -m lapdog.server.

CI status: current GitHub checks pass across check, source/wheel builds, release-note validation, and the Python 3.8-3.13 matrix on Ubuntu, macOS, and Windows; publish-only jobs are skipped as expected.

sabrenner:
i think i will land this for now and double check before releasing if for whatever reason we want to revert this. most Lapdog users should be unaffected as we generally promote CLI usage. 

i believe the only place we reference the Docker implementation is in this repo, which we have updated.

PR #419, opened 2026-09-18
fix(lapdog): codex initial user messages, capturing non-lapdog codex sessions

sabrenner:
Fixes two issues:
- initial codex prompts were not being captured, likely due to outdated code interacting with newer codex versions
- inspecting ongoing codex sessions that were not started with `lapdog`


lapdog/codex_watcher.py
```diff
@@ -337,6 +382,14 @@ def _drain_file(
                 else:
                     state.matches_cwd = _is_under(record_cwd, cwd)
 
+            if session_id and session_ownership is not None:
+                if state.matches_cwd is not True or not session_ownership.accepts(record):
+                    state.ignored = True
+                    state.buffer.clear()
+                    state.buffer_ends.clear()
+                    state.offset = line_end
+                    continue
```
tomshen: [P2] Stop draining a session after rejecting its ownership

`continue` only skips the rejected `session_meta` line. Subsequent records in the same drain have no `session_id` from `_record_session_id`, so they bypass this check and reach `_post_record` with the already-populated `state.session_id` and matching cwd. I reproduced this with a live watcher: after accepting one root, writing an unrelated rollout containing session metadata, a user message, and task completion together still created an unrelated trace with that prompt and output. Return/break when rejecting the file and ensure every drain path honors `state.ignored`; otherwise the common case where several records are written between discovery polls defeats the new isolation behavior.

tomshen (commented):
Found two issues in session isolation and prompt deduplication, detailed inline.

QA used a running Lapdog server and watcher with synthetic Codex JSONL records over HTTP. Initial response-item prompts were captured, injected AGENTS.md content was excluded, the matching legacy user-message event was deduplicated, and an owned child session was captured. The same QA reproduced unrelated-session traces passing the ownership filter and an empty input on a repeated legacy prompt.

lapdog/codex_hooks.py
```diff
@@ -1039,6 +1050,10 @@ def _handle_event_msg(self, session: CodexSession, record: Dict[str, Any]) -> Li
         if event_type == "user_message":
             message = event.get("message", "")
             completed: List[CompletedTrace] = []
+            if message and session.user_prompts and _normalized_text(message) == _normalized_text(
+                session.user_prompts[-1]
+            ):
+                return []
```
tomshen: [P2] Limit prompt deduplication to the same message occurrence

`session.user_prompts` survives `_finalize_turn`, so comparing only text here also discards a legitimate next prompt when no new `task_started`/`turn_context` record precedes it. This is an already-supported legacy flow. Posting `user_message("continue")`, `task_complete`, then another `user_message("continue")` and `task_complete` to the running server produced a second root with an empty input. Repeated prompts during an open turn can also lose the turn boundary. Scope deduplication to the response-item/event pair for the same turn/message rather than the last prompt text; the new response-item branch has the same cross-turn text check.

lapdog/codex_hooks.py
```diff
@@ -1913,7 +1967,7 @@ def _dispatch(
         elif record_type == "event_msg":
             return self._handle_event_msg(session, record)
         elif record_type == "response_item":
-            self._handle_response_item(session, record)
+            return self._handle_response_item(session, record)
```
tomshen, line 1970: [P2] Return an empty list from the assistant deduplication branch

Returning `_handle_response_item` directly now exposes its existing bare `return` at line 1170 to `handle_hook`, which iterates the result as completed traces. When an assistant response item matches the output of an already-emitted LLM span, this yields `None` and the valid hook request fails. I reproduced it on the running server by posting a user message, token usage, `agent_message("Hello.")`, then a response-item assistant message containing `Hello.`: `/codex/hooks` returned HTTP 400 with `'NoneType' object is not iterable`. The watcher retries successfully because the record was already added to `seen_records`, but this still introduces failed hook posts and unnecessary retries. Change that early return to `return []`. The [CI check](https://github.com/DataDog/dd-apm-test-agent/actions/runs/35380251090/job/105714454682) reports the same branch as `lapdog/codex_hooks.py:1170: error: Return value expected`.

tomshen (commented):
Verified both previous findings are fixed: unrelated rollouts remain excluded during capture and shutdown, and repeated prompts preserve separate turns and their inputs. One new runtime regression remains in the response-item return path, detailed inline.

QA used a running Lapdog server and watcher with synthetic Codex JSONL over HTTP. Checked initial prompt capture and injected-context filtering, owned child capture, repeated legacy and response-item prompts, and duplicate prompt pairs in both arrival orders. An assistant response item whose output was already captured returned HTTP 400; retrying that record succeeded.

tomshen (approved):
Verified the return-value fix: the assistant response item that previously failed now returns HTTP 200 on its first delivery and on replay. No remaining actionable findings.

Repeated QA with a running Lapdog server and watcher using synthetic Codex JSONL over HTTP. Initial prompts are captured, injected context is excluded, repeated legacy and response-item prompts retain separate turn inputs, duplicate prompt pairs work in both arrival orders, owned child sessions are captured, and unrelated rollouts stay excluded during capture and shutdown. No runtime errors were observed in these scenarios.

